PR #7, opened 2016-08-16
Simplify context and app parameters

rdumont:
1. The `app` segment of App routes was represented by three parameters (`vendor`, `name` and `version`), one of them optional. Many times all we have is an ID (`vendor.name@version`), which is **exactly** what the API route expects. So, I think that always accepting an ID, either with or without version, makes more sense and cuts one step in most cases, besides looking cleaner.

2. The `context` parameter used to be in the format `vendor.name/vendor.name/vendor.name`. This is how the API expects it, but not how it would be used programmatically. All a client knows about the context is a succession of apps, but not what joins them. Therefore, I believe that accepting an array of apps (`[ 'vendor.name', 'vendor.name', 'vendor.name' ]`) is easier to use.

### DIFF
--- a/src/AppsClient.js
+++ b/src/AppsClient.js
@@ -26,25 +26,25 @@ class AppsClient {
     return this.http.post(url).send(descriptor).thenJson()
   }
 
-  uninstallApp (account, workspace, vendor, name) {
-    checkRequiredParameters({account, workspace, vendor, name})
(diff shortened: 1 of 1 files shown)


firstdoit:
Hm, I guess it's not a problem :) less parsing of locators everywhere! hehe.

PR #8, opened 2016-08-17
Add support for files in VBase client

firstdoit:
This adds support for listing, reading files from and saving files to VBase.

Also fixes the `getAppFile` endpoint so that it returns text instead of json.


rdumont:
Two things:
1. IMO the method names could be improved, so instead of `file`, `files` and `save` they would be called `getFile`, `listFiles` and `saveFile`.
2. No `deleteFile`? :(

firstdoit:
@rdumont fair enough! Fixed.

PR #9, opened 2016-08-18
Support app settings

rdumont:
This implements two new operations: get and patch settings. The patch API is implemented according to the [RFC 7396 - JSON Merge Patch](https://tools.ietf.org/html/rfc7396) specification.

The `PUT` settings operation has also been updated to use its canonical endpoint (`:account/:workspace/apps/:app/settings`).

Furthermore, query string parameters are now only being passed when they are not empty or undefined.

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -55,3 +55,15 @@ Request.prototype.thenJson = function () {
 Request.prototype.thenText = function () {
   return this.then(res => res.text())
 }
+
+Request.prototype.baseQuery = Request.prototype.query
+Request.prototype.query = function (options) {
+  let query = {}
+  for (let key of Object.keys(options)) {
+    let value = options[key]
+    if (value !== undefined && value !== '') {
+      query[key] = value
+    }
+  }
+  return this.baseQuery(query)
+}
--- a/src/http.js
+++ b/src/http.js
@@ -56,7 +56,7 @@ Request.prototype.thenText = function () {
   return this.then(res => res.text())
 }
 
-Request.prototype.baseQuery = Request.prototype.query
+const baseQuery = Request.prototype.query
 Request.prototype.query = function (options) {
   let query = {}
   for (let key of Object.keys(options)) {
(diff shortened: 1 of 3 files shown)


src/http.js
```diff
@@ -55,3 +55,14 @@ Request.prototype.thenJson = function () {
 Request.prototype.thenText = function () {
   return this.then(res => res.text())
 }
+
+Request.prototype.options = function (options) {
```
firstdoit: Why call this `options`? I'd rather monkeypatch `query` completely, like we do with `then`. That would make client code more obvious. Or at least something `definedQuery`?

firstdoit:
One comment, then LGTM

src/http.js
rdumont: Sure, I think I'll go with the monkeypatch.

rdumont:
LGTY?

rdumont:
I accidentally implemented an endless recursion in `http.query`. Oops. Fixed.

src/http.js
```diff
@@ -55,3 +55,15 @@ Request.prototype.thenJson = function () {
 Request.prototype.thenText = function () {
   return this.then(res => res.text())
 }
+
+Request.prototype.baseQuery = Request.prototype.query
```
firstdoit: I suggest you use `const prototypeQuery = Request.prototype.query` (as in line 23) to avoid a circular assignment if another project does something similar. We ran into this error ourselves, which made me use a local `const`. :)

firstdoit:
One comment, then LGTM! 😂

src/http.js
```diff
@@ -65,5 +65,5 @@ Request.prototype.query = function (options) {
       query[key] = value
     }
   }
-  return this.baseQuery(query)
+  return baseQuery.apply(this, [query])
```
firstdoit: Fun fact: if you use `call` you don't need to create a new Array to pass the arguments in. Which would matter if we were writing a space ship! #microoptimizations

firstdoit:
LGTM!

PR #11, opened 2016-10-03
Reimplement using axios and add flow typing

firstdoit:
- [x] Gets rid of our dreaded monkey-patched `http` based on `requisition`, instead using `axios`.
- [x] Changes the API for client construction (argument list instead of object).
- [x] Changes `RegistryClient.publishApp` to use a `multipart/mixed` request based on Vinyl files.
- [x] Changes `RegistryClient.publishApp` query string option from `isPreRelease` to `isDevelopment`.
- [x] Adds flow typing.
- [x] Adds tests for any endpoints which are more complex than GET/PUT/POST/DELETE/PATCH with query string params.


tamorim:
@firstdoit I actually find the object model more developer friendly than the argument list.
With an argument list you make the developer have a cognitive load to remember the order of the parameters.
With an object he only needs to remember the config keys and the order doesn't matter.

src/RegistryClient.js
```diff
@@ -1,123 +1,106 @@
...
+   */
+  publishApp (account: string, workspace: string, files: Array<File>, isDevelopment?: boolean = false) {
+    if (!(files[0] && files[0].path && files[0].contents)) {
+      throw new Error('Argument files must be an array of {path, contents}, where contents can be a String, a Buffer or a ReadableStream.')
+    }
+    const indexOfManifest = files.findIndex(({path}) => path === 'manifest.json')
+    if (indexOfManifest === -1) {
+      throw new Error('No manifest.json file found in files.')
+    }
+    const sortedFiles = files.splice(indexOfManifest, 1).concat(files)
```
tamorim: @firstdoit Can you please explain this line?
From what I understand it actually does nothing.
When using `splice` you **mutate** the `files` array and then `concat` the return of it, the `manifest` file, with the now mutate `files`.
On the end you have the same array you had at the beginning, no?

tamorim: Oh, I think I got it now... it's just to move the `manifest` file do the head of the array?

tamorim:
@firstdoit LGTM! :+1:

src/RegistryClient.js
firstdoit: Precisely! We send the manifest as the first file in the streaming request so our Apps service can decide early whether it wants to reject the request (e.g. invalid manifest).

firstdoit:
@tamorim I agree that the object is *easier*, but I also think that options objects are not the best way to pass in *required arguments*. I believe this makes it more explicit that these are all required parameters. In the end, you could argue that remembering the order or the keys is equally hard, couldn't you? :)

tamorim:
@firstdoit I wouldn't argue that. I believe that remembering order is way harder than remember the name of something.
Names can be logical or self explanatory, while order (at least in this case) doesn't mean nothing.

firstdoit:
That's just because *you're not a robot*. 🤖

PR #14, opened 2016-12-30
Implements new link's logic

ewilazarus:


ewilazarus:
I removed msg-pack usage for now. It was too troublesome and we have to come up with the MVP ASAP. I'll, therefore, open an issue to remind us to adopt msg-pack in the future

PR #17, opened 2017-03-28
Add a colossus client

mairatma:
`render-builder` currently contains code to access Colossus that is very similar to other clients here in `node-vtex-api`, as can be seen [here](https://github.com/vtex/render-builder/blob/master/service/colossus/index.ts).

I currently need to use the same code in another app (`masterdata-graphql`), so it's best to move that logic here and have it be shared with anyone that needs it too.

I've tested it by manually running the client in node's terminal and making `sendLog` and `sendEvent` requests. It all seems to be working as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 ### SublimeText ###
 *.sublime-workspace
 
+### VSCode ###
+.vscode
+
 ### OSX ###
 .DS_Store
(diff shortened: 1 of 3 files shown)


mairatma:
LOL!! Sorry!

mairatma:
Fixed now!

PR #22, opened 2017-07-25
Add cache with revalidation by etag

brunoabreu:
The objective is not to implement full HTTP Cache specification as described in [RFC 7234](https://tools.ietf.org/html/rfc7234)

It is just an use case of HTTP Cache, and the only one that matters right now, as our infrastructure is based on cache revalidation.

### DIFF
--- a/src/HttpClient/cache.ts
+++ b/src/HttpClient/cache.ts
@@ -1,0 +1,52 @@
+import {AxiosInstance, AxiosResponse, AxiosRequestConfig} from 'axios'
+
+const EMPTY = {}
+
+const successOrNotModified = (status: number): boolean =>
+  status >= 200 && status < 300 || status === 304
+
+export const addCacheInterceptors = (http: AxiosInstance, cacheStorage: CacheStorage) => {
+  http.interceptors.request.use((config: CacheableRequestConfig) => {
+    if (config.cacheable) {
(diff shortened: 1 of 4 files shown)


src/HttpClient/cache.ts
```diff
@@ -0,0 +1,52 @@
...
+  })
+
+  http.interceptors.response.use((response: CacheableResponse) => {
+    const {status, data, headers, config} = response
+    if (status === 304) {
+      return config.cached
+    }
+
+    if (headers.etag) {
+      cacheStorage.set(config.url, {
```
firstdoit: Hmm if I'm not mistaken this means we will cache responses regardless of the method used? That might not be correct: POST and GET might get different responses for the same url.

brunoabreu: We need to explicitly make the request cacheable, as I did [here](https://github.com/vtex/node-vtex-api/blob/3f1dc5f58aac6a56e98ff2dfa69820198db3734a/src/HttpClient/index.ts#L64).

I prefer this way instead of just cache every `GET` because, at least for now, I don't want us to cache buffers (vbase/apps/registry files) and streams (gzipped bundles), but only API responses. I'm concerned about memory usage. We will get there in the next caching iteration.

firstdoit: Perfect!

PR #23, opened 2017-07-25
Allow specifying options when calling VBase

igorframos:


src/VBase.ts
```diff
@@ -12,6 +12,10 @@ const routes = {
   File: (bucket: string, path: string) => `${routes.Bucket(bucket)}/files/${path}`,
 }
 
+const isVBaseOptions = (opts?: string | VBaseOptions): opts is VBaseOptions => {
+  return (<VBaseOptions>opts) !== undefined
```
brunoabreu: It will not work. TypeScript does not check anything at runtime. Here you need to check some prop of `opts` to make sure it is a `VBaseOptions`, like `prefix`. Unfortunately all props are nullable and if someone call this method with `opts` = `{}`, will be recognized as a `string`.

PR #28, opened 2017-09-14
Remove unnecessary console.log

samuelmtimbo:
An API client library should not console.log errors. This is the reason behind these non user friendly messages on toobelt:

screen shot 2017-09-14 at 13 08 17


firstdoit:
@samuelmtimbo this LGTM but can you implement this message under the `verbose` flag in toolbelt so we still know which call failed? cc @brunoabreu

PR #29, opened 2017-10-03
New deprecateApp command

mrkmat:
New deprecateApp command to be used in toolbelt.
More info will be posted [here](https://github.com/vtex/toolbelt)

### DIFF
--- a/src/Registry.ts
+++ b/src/Registry.ts
@@ -50,6 +50,10 @@ export class Registry {
     return this.http.get<RegistryAppVersionsList>(routes.App(app))
   }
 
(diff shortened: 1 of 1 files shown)


src/Registry.ts
```diff
@@ -50,6 +50,10 @@ export class Registry {
     return this.http.get<RegistryAppVersionsList>(routes.App(app))
   }
 
+  deprecateApp = (app: string, version: string, deprecated: boolean) => {
```
igorframos: The name of the command is a bit strange. If you are deprecating the app as the command name suggests, why would you need to pass the value of `deprecated`?

PR #30, opened 2017-10-30
Updating logs and events sender to colossus

iagoaraujo:


src/Colossus.ts
```diff
@@ -12,11 +16,11 @@ export class Colossus {
     this.http = HttpClient.forWorkspace('colossus', opts)
   }
 
-  sendLog = (sender: string, subject: string, message: any, level: string) => {
-    return this.http.post(routes.Log(sender, subject, level), message)
+  sendLog = (subject: string, message: any, level: string) => {
+    return this.http.put(routes.Log(subject, level), message)
```
brunoabreu: There is a better way to add querystring parameters when using Axios. Take a look [here](https://github.com/axios/axios#request-config).

Basically there is a third parameter, the request config. The `params` prop is encoded and sent as querystring.

```
return this.http.put(routes.Log(subject, level), message, {params: {subject}})
```

PR #31, opened 2017-11-14
Add disk cache

amoreira:
Uses disk cache to improve responses of buffers and streams methods.
The most affected resources are `Apps` and `Registry`.

You can use the gist bellow to test:
https://gist.github.com/amoreira/3ceef0fe69650d905f9e16baa39edacd

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,10 +31,12 @@
     "tar-fs": "^1.15.3"
   },
   "devDependencies": {
+    "@types/fs-extra": "^4.0.4",
     "@types/archiver": "^2.0.0",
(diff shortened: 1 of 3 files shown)


firstdoit:
I've discussed this with @brunoabreu and there are two problems to the decision of caching in this layer:

1 - Render will no longer serve published apps files, that will be the apps system responsibility directly
2 - During SSR, we fetch files that do not come from apps, such as `unpkg` libs. This cache would not help there.

So we've decided to add the file cache layer directly to render-server and I'm dropping this PR.

PR #34, opened 2017-11-27
Updating metadata and vbase with new security API

iagoaraujo:


src/Metadata.ts
```diff
@@ -2,9 +2,9 @@ import { HttpClient, InstanceOptions } from './HttpClient'
 import { BucketMetadata } from './responses'
 
 const routes = {
-  Bucket: (bucket: string) => `/buckets/${bucket}`,
-  Metadata: (bucket: string) => `/buckets/${bucket}/metadata`,
-  MetadataKey: (bucket: string, key: string) => `/buckets/${bucket}/metadata/${key}`,
+  Bucket: (bucket: string) => `/buckets/${process.env.VTEX_APP_NAME}/${bucket}`,
```
brunoabreu: This environment variable does not include the app _vendor_. It is a problem, right? What if I have `vtex.banana` and `omnilogic.banana`?

And I think we should throw an error in `Metadata` and `VBase` constructors if that variable is not available, just in case.

iagoaraujo: Yeah, I should be using `VTEX_APP_ID`.

src/Metadata.ts
```diff
@@ -2,7 +2,12 @@ import { HttpClient, InstanceOptions } from './HttpClient'
 import { BucketMetadata } from './responses'
 
 const routes = {
-  Bucket: (bucket: string) => `/buckets/${process.env.VTEX_APP_NAME}/${bucket}`,
+  Bucket: (bucket: string) => {
+    if (!process.env.VTEX_APP_ID) {
```
brunoabreu: `VTEX_APP_ID` includes app version, then `vtex.render-server@5.0.12` and `vtex.render-server@5.0.13` will consume different buckets, or VBase/Metadata are handling that?

src/Metadata.ts
```diff
@@ -2,7 +2,12 @@ import { HttpClient, InstanceOptions } from './HttpClient'
...
+      throw new Error(`Invalid path to access Metadata. Variable VTEX_APP_ID is not available.`)
```
brunoabreu: I think we should keep it clean and raise that error in constructor.

It would be nice to create a `const appId = process.env.VTEX_APP_ID` at the begging of this module and use it instead of  `process.env.VTEX_APP_ID` everywhere.

PR #47, opened 2018-07-20
Return bundle size with link and publish errors

iagoaraujo:


src/Apps.ts
```diff
@@ -86,10 +86,16 @@ export class Apps {
...
     const finalize = zip.finalize()
-
-    const [response] = await Promise.all([request, finalize])
-    response.bundleSize = zip.pointer()
-    return response
+    const bundleSize = zip.pointer()
+
+    try {
+      const [response] = await Promise.all([request, finalize])
+      response.bundleSize = zip.pointer()
```
grovina: `bundleSize` instead of `zip.pointer()`?

PR #50, opened 2018-07-30
Add option to return null on not found

brunoabreu:
It also adds new vbase methods `getJSON` and `saveJSON`. Without that we can't have cached responses for JSON files because getBuffer and getStream responses are not cached.

### DIFF
--- a/src/HttpClient/cache.ts
+++ b/src/HttpClient/cache.ts
@@ -3,11 +3,11 @@ import {URL, URLSearchParams} from 'url'
 
 const EMPTY = {etag: null, response: null}
 
-const successOrNotModified = (status: number): boolean =>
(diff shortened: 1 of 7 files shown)


src/HttpClient/cache.ts
```diff
@@ -3,11 +3,11 @@ import {URL, URLSearchParams} from 'url'
 
 const EMPTY = {etag: null, response: null}
 
-const successOrNotModified = (status: number): boolean =>
-  status >= 200 && status < 300 || status === 304
+const addNotModified = (validateStatus: (status: number) => boolean) =>
```
firstdoit: A thing of beauty

firstdoit:
👀 @victorges

PR #60, opened 2018-09-03
Add max-age caching support

brunoabreu:
This is a naive implementation while we are looking for a new HTTP request library with native caching support.

It's not possible to interrupt an axios request and respond from cache using interceptors. So this PR drops almost all interceptors in favor of koa-style middlewares. That will also make our future migration away from axios easier.


src/HttpClient/middlewares/cache.ts
```diff
@@ -0,0 +1,105 @@
...
+    await next()
+
+    const {response} = ctx
+    if (!response) {
+      return
+    }
+
+    if (response.status === 304) {
+      ctx.response = {...cached!.response, status: 304}
+      return
```
firstdoit: So if we expire but revalidate with etag, we never re-add time to the cache key? It will only be 304's from now on

src/HttpClient/middlewares/cache.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    const {headers} = response
+    const {age, etag, maxAge, noStore} = parseCacheHeaders(headers)
+    if (noStore && !etag) {
```
firstdoit: For the future, this should be only `noStore` when vbase stops sending this incorrectly

src/HttpClient/middlewares/notFound.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  await next()
+}
+
+export const notFoundFallbackMiddleware = async (ctx: MiddlewareContext, next: () => Promise<void>) => {
+  await next()
+
+  const {config} = ctx
+  if (nullIfNotFound(config) && ctx.response) {
```
firstdoit: Shouldn't we check if the response is 404?

PR #61, opened 2018-09-04
Add LRU cache implementation

pedroig:
This PR adds a new client that provides a wrapper class to the LRU Cache. The purpose of this wrapper is to add a `getStats` method that provides statics on cache usage.


pedroig:
Looks like the CI failed because of the package dependencies that I've added. How should I add new packages? Do I need to do something different than `yarn add {package-name}`?

PR #62, opened 2018-09-06
LRUCache - revised metrics interface

pedroig:
Minor change in the metrics interface for a better compatibility with previous use cases of the npm package `lru-cache` at [render-server](https://github.com/vtex/render-server).

### DIFF
--- a/src/LRUCache.ts
+++ b/src/LRUCache.ts
@@ -31,8 +31,9 @@ export class LRUCache <K, V> {
 
   public getStats = () => {
     const stats = {
-      count: this.storage.itemCount,
-      disposed: this.disposed,
+      itemCount: this.storage.itemCount,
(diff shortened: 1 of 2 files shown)


pedroig:
I am skipping the 0.45.1 version on purpose because I have wrongly published this version in a beta npm tag. However, versions cannot be overwritten not even with a different tag. I have even unpublished this version, but it is still not possible to use it again, the only and unpleasant solution is to skip 0.45.1.

src/LRUCache.ts
```diff
@@ -31,8 +31,9 @@ export class LRUCache <K, V> {
 
   public getStats = () => {
```
tlgimenes: It would be amazing if you also add typings to the stats response

PR #63, opened 2018-09-13
Added the client AppIds

pedroig:
This PR adds a new client to the API that aims to create a standard for the operations that deal with AppIds.


igorframos:
I agree with @firstdoit. It does look a lot more like `AppId` stuff than `Semver`.

pedroig:
Agreed, `AppId` wins!

PR #66, opened 2018-09-20
Add accept-encoding gzip

brunoabreu:


src/HttpClient/index.ts
```diff
@@ -38,6 +38,7 @@ export class HttpClient {
   private constructor (opts: ClientOptions) {
     const {baseURL, authToken, authType, cacheStorage, recorder, userAgent, timeout = DEFAULT_TIMEOUT_MS} = opts
     const headers = {
+      'Accept-Encoding': 'gzip',
```
victorges: Will it ungzip automatically if **we** set it? Weird API if so but I guess it's the way to do it

firstdoit: I think they are orthogonal, they ungzip if the origin replies with a `Content-Encoding`. The origin doesn't necessarily respects your `Accept-Encoding`. So... make sense. #nicetry #gohomego #nodeftw #15yodiscussions #flamewars

brunoabreu: @firstdoit it right

PR #68, opened 2018-09-27
Feature/Extendable MetricsAccumulator

pedroig:
This PR implements changes in the MetricsAccumulator class aiming for a more extendable version of it. In addition, there are two extra changes:

- the removal of aggregated metrics with a low maximum value;

- the addition of the minimum as an aggregation property.

### DIFF
--- a/src/MetricsAccumulator.ts
+++ b/src/MetricsAccumulator.ts
@@ -89,7 +89,7 @@ export class MetricsAccumulator {
     return this.flushMetrics()
   }
 
-  protected createMetricToAggregateReducer = (production: boolean) => (value: any, key: string, obj: any): EnrichedAggregateMetric => { //must be protected in order to override
+  protected createMetricToAggregateReducer = (production: boolean) => (value: any, key: string, obj: any): EnrichedAggregateMetric => {
     const metricToAggregateReducer: EnrichedAggregateMetric = {
       name: key,
       production,
@@ -132,7 +132,7 @@ export class MetricsAccumulator {
     return [...aggregateMetrics, ...aggregateDevMetrics]
   }
 
-  protected flushMetrics = (): Metric[] => {      // must it be protected in order to override
+  protected flushMetrics = (): Metric[] => {
 
     const systemMetrics: Metric[] = [
       {


src/MetricsAccumulator.ts
```diff
@@ -104,21 +89,50 @@ export class MetricsAccumulator {
...
-  private flushMetrics = (): Metric[] => {
-    const aggregateMetrics: Metric[] = values(mapObjIndexed(
-      this.metricToAggregate,
+  protected resetAccumulators = () => {
+    this.metricsMillis = {}
+    this.devMetricsMillis = {}
+  }
+
+  protected getAggregateMetrics = (): Metric[] => {
+    const removeIrrelevantMetrics = (metrics: EnrichedAggregateMetric[]): Metric[] =>
```
firstdoit: I _strongly_ disagree with this. We have no idea whether `10` is "irrelevant" for a given arbitrary metric.

src/MetricsAccumulator.ts
```diff
@@ -104,21 +89,50 @@ export class MetricsAccumulator {
     return this.flushMetrics()
   }
 
-  private cacheToMetric = (value: GetStats, key: string): Metric => ({
-    name: `${key}-cache`,
-    ...value.getStats(),
+  protected createMetricToAggregateReducer = (production: boolean) => (value: any, key: string, obj: any): EnrichedAggregateMetric => {
```
firstdoit: Why is this protected? Will it be called by a subclass?

src/MetricsAccumulator.ts
```diff
@@ -104,21 +89,50 @@ export class MetricsAccumulator {
...
+    const metricToAggregateReducer: EnrichedAggregateMetric = {
+      name: key,
+      production,
+      ...this.aggregate(value),
+    }
+    return metricToAggregateReducer
+  }
+  
+  protected aggregate = (value: number[]): AggregateMetric => ({
```
firstdoit: Is this the part you intend to override in a subclass method? I think we shouldn't. It would be preferable to have all metrics export the same basic keys, and potentially *add more keys*. Or do you think some of these might be irrelevant for some types of metrics? @tlgimenes

src/MetricsAccumulator.ts
```diff
@@ -104,21 +89,50 @@ export class MetricsAccumulator {
...
+    this.metricsMillis = {}
+    this.devMetricsMillis = {}
+  }
+
+  protected getAggregateMetrics = (): Metric[] => {
+    const removeIrrelevantMetrics = (metrics: EnrichedAggregateMetric[]): Metric[] => 
+      filter(metric => metric.max > 10, metrics)
+
+    const aggregateMetricsAll: EnrichedAggregateMetric[] = values(mapObjIndexed(
+      this.createMetricToAggregateReducer(true),
```
firstdoit: This is causing the creating of a new function every time `statusTrack` is called. Do we need to regenerate it always?

src/MetricsAccumulator.ts
```diff
@@ -8,13 +8,17 @@ interface Metric {
...
+  max: number
   mean: number
   median: number
+  min: number
   percentile95: number
   percentile99: number
-  max: number
+}
+
+interface EnrichedAggregateMetric extends Metric, AggregateMetric {
```
firstdoit: I would rather have a smaller name 😂

pedroig: what about `ContextMetric` ?

src/MetricsAccumulator.ts
pedroig: no, this is the case that I intend to override

src/MetricsAccumulator.ts
pedroig: no, this is `protected` because I need access to it in the subclass, I am not planning to override it

src/MetricsAccumulator.ts
pedroig: yes, I wasn't completely sure if this would be applicable in the general case, I'll remove it then

src/MetricsAccumulator.ts
pedroig: You are right, I'll do it the way it was before

PR #69, opened 2018-09-28
Colossus refactor and ordering logs

athoscouto:
Refactor Colossus into Logger, placing billing metrics in Billing and sending events in Event. Added queue to manage Colossus logging in a ordered fashion on builder hub.

It solves the problem of logs arriving at Colossus in a different order that they were created. It also removes the legacy Colossus object in more specialised classes. 

Added commands to log numbers from 1 to 10 in order on `admin-iframe`. Logged it with and without the changes.

#### Screenshots of manual test
Before adding the queue:
screen shot 2018-09-28 at 15 41 10

After adding the queue:
screen shot 2018-09-28 at 15 44 33

#### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Requires change to documentation, which has been updated accordingly.


package.json
```diff
@@ -24,6 +24,8 @@
   "license": "MIT",
   "dependencies": {
     "@types/lru-cache": "^4.1.1",
+    "@types/p-queue": "^2.3.1",
+    "@types/ramda": "^0.25.38",
```
firstdoit: 👀

PR #72, opened 2018-10-06
Multilayered Cache Client

pedroig:
This PR adds a new client that provides multilayered cache support built on top of general cache layers that follow the basic cache interface (get, set, has). In addition, the client `DiskCache` is also added as an alternative cache storage that can be used as a layer for the `MultilayredCache`.

**Breaking change:** The following operations on the `LRUCache` class are now asynchronous: get, has, set. This was necessary in order to follow a standard interface for each possible cache layer.


src/caches/MultilayeredCache.ts
```diff
@@ -0,0 +1,88 @@
...
+  public get = async (key: K, fetcher?: () => V): Promise<V | void> => {
+    let value: V | void
+    let successIndex = await this.findIndex(async (cache: CacheLayer<K, V>) => {
+      const [getValue, hasKey] = await Promise.all([cache.get(key), cache.has(key)])
+      value = getValue
+      return hasKey
+    }, this.caches)
+    if (successIndex === -1) {
+      if (fetcher) {
+        value = fetcher()
```
pedroig: must be async

src/caches/DiskCache.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  private hits = 0
+  private total = 0
+
+  constructor(private cachePath: string, private readFile=readJSON, private writeFile=outputJSON) {}
+
+  public has = async (key: string): Promise<boolean> => {
+    const cacheKey = this.getCacheKey(key)
+    try {
+      await stat(cacheKey)
```
brunoabreu: We could call `pathExists` from `fs-extra`. It calls `fs.access` and already implements this try/catch logic.

> To check if a file exists without manipulating it afterwards, fs.access() is recommended.
https://nodejs.org/api/fs.html#fs_fs_stat_path_options_callback

src/caches/MultilayeredCache.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+    const failedCaches = slice(0, successIndex, this.caches)
+    failedCaches.forEach(cache => cache.set(key, value as V))
+    return value
+  }
+
+  public set = async (key: K, value: V) => {
+    let setInAtLeastOneCache = false
+    this.caches.forEach(async (cache: CacheLayer<K, V>) => {
+      setInAtLeastOneCache = setInAtLeastOneCache || await cache.set(key, value)
```
brunoabreu: This will write only on the first cache layer, because when `setInAtLeastOneCache=true` there will be a short-circuit and `cache.set` will never be called.

Take a look here: https://codeburst.io/javascript-short-circuit-evaluation-3709ffda6384

PR #74, opened 2018-10-22
Add exponential backoff for the HttpClient

pedroig:
This PR adds a retry delay for the `HttpClient`. 

Before no delay was applied between retries, now the retry attempts follow an exponential backoff.

#### Types of changes

### DIFF
--- a/src/HttpClient/index.ts
+++ b/src/HttpClient/index.ts
@@ -11,7 +11,7 @@ import {CacheableRequestConfig, Cached, cacheMiddleware} from './middlewares/cac
 import {metricsMiddleware} from './middlewares/metrics'
 import {acceptNotFoundMiddleware, notFoundFallbackMiddleware} from './middlewares/notFound'
 import {Recorder, recorderMiddleware} from './middlewares/recorder'
-import {defaultsMiddleware, requestMiddleware} from './middlewares/request'
+import {defaultsMiddleware, getAxiosInstance, requestMiddleware} from './middlewares/request'
 
 const DEFAULT_TIMEOUT_MS = 10000
 const noTransforms = [(data: any) => data]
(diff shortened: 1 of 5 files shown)


src/HttpClient/middlewares/request.ts
```diff
@@ -11,7 +11,10 @@ const http = axios.create({
   }),
 })
 
-retry(http)
+retry(http, {
+  retries: 3,
```
firstdoit: This should be configurable by the user

src/HttpClient/index.ts
```diff
@@ -60,7 +60,7 @@ export class HttpClient {
   private runMiddlewares: compose.ComposedMiddleware<MiddlewareContext>
 
   public constructor (opts: ClientOptions) {
-    const {baseURL, authToken, authType, cacheStorage, metrics, recorder, userAgent, timeout = DEFAULT_TIMEOUT_MS} = opts
+    const {baseURL, authToken, authType, cacheStorage, metrics, recorder, retries = 3, userAgent, timeout = DEFAULT_TIMEOUT_MS} = opts
```
firstdoit: @brunoabreu what do you think? should retries be default? cc @tlgimenes

src/HttpClient/middlewares/request.ts
```diff
@@ -1,29 +1,34 @@
...
+export const getAxiosInstance = (retries: number) => {
+  const http = axios.create({
+    httpAgent: new Agent({
+      keepAlive: true,
+      maxFreeSockets: 50,
+    }),
+  })
+
+  retry(http, {
+    retries,
```
firstdoit: Should we even call `retry` if retries is 0?

src/HttpClient/index.ts
pedroig: 3 retries were the previous default definition, I just kept as it used to be in spite of the exponential backoff addition.

src/HttpClient/middlewares/request.ts
pedroig: If 0 is passed, the behaviour is the desired one. It would add a useless interceptor, but the overhead is not meaningful.

PR #85, opened 2019-02-13
HttpClient with cache varying with segment

vwraposo:
This PR makes the HttpClient cache vary according to the segment (if the response varies)

Responses that varies with the segment were being cached without the segment information, therefore when the segment changed the client returned the cached data, even though it was incorrect.

#### Types of changes

### DIFF
--- a/src/HttpClient/middlewares/cache.ts
+++ b/src/HttpClient/middlewares/cache.ts
@@ -1,5 +1,5 @@
 import {AxiosRequestConfig, AxiosResponse} from 'axios'
-import {contains} from 'ramda'
+import {match} from 'ramda'
 import {URL, URLSearchParams} from 'url'
 import {CacheLayer} from '../../caches/CacheLayer'
 import {MiddlewareContext} from '../context'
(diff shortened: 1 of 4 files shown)


src/HttpClient/middlewares/cache.ts
```diff
@@ -36,14 +36,17 @@ function isCacheable (arg: any): arg is CacheableRequestConfig {
...
-export const cacheMiddleware = (cacheStorage: CacheLayer<string, Cached>) => {
+export const cacheMiddleware = ({cacheStorage, segmentToken}: {cacheStorage: CacheLayer<string, Cached>, segmentToken: string}) => {
   return async (ctx: MiddlewareContext, next: () => Promise<void>) => {
     if (!isCacheable(ctx.config)) {
       return await next()
     }
 
     const key = cacheKey(ctx.config)
-    const cached = await cacheStorage.get(key)
+    const keyWithSegment = key + segmentToken
```
tlgimenes: why not interpolating the string for better understanding ?

tlgimenes:
```suggestion
    const keyWithSegment = `${key}${segmentToken}`
```

PR #115, opened 2019-04-04
Feature/export error logger

diegoximenes:
Exports a `logError` function. 
Useful to log errors thrown by functions executed asynchronously with the same format used by service.

#### Types of changes

### DIFF
--- a/src/utils/error.ts
+++ b/src/utils/error.ts
@@ -1,5 +1,7 @@
 // Inspired by https://github.com/sindresorhus/serialize-error
 import { find, keys, pick } from 'ramda'
+import { IOClients } from '../clients/IOClients'
+import { ServiceContext } from '../service/typings'
 
 export const PICKED_AXIOS_PROPS = ['baseURL', 'cacheable', 'data', 'finished', 'headers', 'method', 'timeout', 'status', 'path', 'url']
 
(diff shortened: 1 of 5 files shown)


src/utils/error.ts
```diff
@@ -111,3 +113,51 @@ export const cleanError = (value: any) => {
...
+
+/**
+ * Send error through the logger.
+ *
+ * @param ctx service context related to the error
+ * @param error an object representing the error
+ */
+export function logError<T extends IOClients, U, V> (ctx: ServiceContext<T, U, V>, error: any) {
+  console.error('[node-vtex-api error]', error)
+  const cleanedError = cleanError(error)
```
firstdoit: This will cause error to be cleaned twice, since this is done here: https://github.com/vtex/node-vtex-api/pull/115/files#diff-139e2f0fc6a2190b653bea26e664af49R17

diegoximenes: @firstdoit 
One example:
```
async function makeReq () {
  ...
  await axios.request(...)
  ...
}
makeReq().catch(makeReqError => logError(ctx, makeReqError))
```

Without this PR I can't easily log makeReqError with the same format (operationId, etc) of the errors "synchronously" detected in the request flow.

diegoximenes: I thought that cleanError(cleanError(error)) == cleanError(error), or at least their differences didn't matter. I also considered that cleanError is cheap to compute. I will take I better look on that.

firstdoit: It is, and they are. 

However I don't think it's idiomatic to use a `logError` if we have a `clients.logger.error()` available. Why not add the `ctx` there instead?

diegoximenes: Done, I kept duplicate cleanError call though.

firstdoit:
@diegoximenes do you still need this? I tried to understand our discussion but couldn't.

diegoximenes:
@firstdoit I don't need because I have implemented this kind of functionality above "`@vtex/api` layer" (this was before the existence of `Service` and etc).
This functionality will be useful to log exceptions that are thrown in an async context outside the main income request flow, e.g.: https://github.com/vtex/node-vtex-api/pull/115#discussion_r275829124

This PR needs a rebase though.

PR #124, opened 2019-04-15
Feature/socket stats

diegoximenes:

- Tracks number of sockets per origin.
- Tracks number of requests not assigned to a socket per origin.

These metrics will be useful to fine tune `maxFreeSockets`, `maxSockets`, and the maximum number of sockets per node process.
Currently `maxSockets` is set to infinity.

#### Types of changes

### DIFF
--- a/src/service/utils/incomingRequestStats.ts
+++ b/src/service/utils/incomingRequestStats.ts
@@ -1,0 +1,29 @@
+import { IOClients } from '../../clients/IOClients'
+import { ServiceContext } from '../typings'
+
+class IncomingRequestStats {
+  public closed = 0
+  public total = 0
+
+  public get () {
+    return {
+      closed: this.closed,
+      total: this.total,
+    }
+  }
+
+  public clear () {
+    this.closed = 0
+    this.total = 0
+  }
+}
+
+export const incomingRequestStats = new IncomingRequestStats()
+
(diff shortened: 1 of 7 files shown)


firstdoit:
Rebase and let's put it on service-runtime-node asap :)

firstdoit:
Oh! There's one more thing you could track. We need to know how many cancelled requests we respond nowadays to inform how important it is to have cancellation token implemented.

https://github.com/koajs/koa/issues/1088

Basically, in one of the first middlewares like timings, accumulate a counter for each `close` event. 
```
ctx.req.on('close', () => {
  // your logic here...
});```

Remember to re-use the same function (declared once, outside the middleware). 

What do you think, @diegoximenes ?

File: src/service/utils/incomingRequestStats.ts [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+  public clear () {
+    this.closed = 0
+    this.total = 0
+  }
+}
+
+export const incomingRequestStats = new IncomingRequestStats()
+
+export async function trackIncomingRequestStats <T extends IOClients, U, V> (ctx: ServiceContext<T, U, V>, next: () => Promise<any>) {
+  ctx.req.on('close', () => {
```
firstdoit: You don't need to redeclare this function on every handler, it could be declared outside this scope

diegoximenes: done

PR #132, opened 2019-04-23
Implement IOClient

firstdoit:

We decided to deprecate IODataSource a long time ago. The time has come.

DataSources are a **Terrible Idea®**.

#### Types of changes


src/HttpClient/HttpClient.ts
```diff
@@ -0,0 +1,137 @@
...
+    const postConfig: RequestConfig = {...config, url, data, method: 'post'}
+    return this.request(postConfig) as Promise<IOResponse<T>>
+  }
+
+  public patch = <T = void>(url: string, data?: any, config: RequestConfig = {}): Promise<T> => {
+    const patchConfig: RequestConfig = {...config, url, data, method: 'patch'}
+    return this.request(patchConfig).then(response => response.data as T)
+  }
+
+  public delete = (url: string, config?: RequestConfig): Promise<IOResponse<void>> => {
```
marcosvcp: Why delete does not receive a generic type definition like others? I think it should

firstdoit: Delete shouldn't have a response, right? Anyway, didn't change this bit.

marcosvcp: Nope, delete should include a deleted-entity if response was 200

kevinch: Agreed with Marcos, i think it should return the deleted element.

src/HttpClient/AppClient.ts
```diff
@@ -0,0 +1,25 @@
+import { IOContext } from '../service/typings'
+
+import { IOClient } from './IOClient'
+import { AuthType, InstanceOptions } from './typings'
+
+/**
+ * Used to perform calls on apps you declared a dependency for in your manifest.
+ */
+export class AppClient extends IOClient {
```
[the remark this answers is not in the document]
firstdoit: Wow, complicated question. I'm not sure. One option is to have a separate CrossAccountAppClient. Another option is to indicate in the params...

src/HttpClient/AppClient.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(app: string, context: IOContext, options: InstanceOptions) {
```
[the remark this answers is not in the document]
firstdoit: Disagree. You should only create requests for apps you depend, and you can only depend on one version. Hence, `app` is necessarily `vendor.name`.

src/IODataSource.ts
```diff
@@ -10,6 +10,9 @@ interface HttpClientFactoryOptions {
 
 export type HttpClientFactory = (opts: HttpClientFactoryOptions) => HttpClient | void
 
+/**
+ * @deprecated Extend one of the subclasses of HttpClient instead.
```
tlgimenes:
```suggestion
 * @deprecated Extend one of the subclasses of IOClient instead.
```

src/clients/LicenseManager.ts
```diff
@@ -0,0 +1,40 @@
...
+      metric: 'lm-account-data',
+    })
+  }
+
+  public getTopbarData (VtexIdclientAutCookie: string) {
+    return this.http.get(routes.topbarData, {
+      headers: {
+        VtexIdclientAutCookie,
+      },
+      memoizeable: false,
```
tlgimenes:
```suggestion
      memoizeable: true,
```

PR #135, opened 2019-04-29
add data param to http cliente delete

jgfidelis:
as title says

I am a bit worried that this may break whoever uses this.delete(url, config), because config will then be the `data`. But making data be the third param would be different with all other methods.

#### Types of changes


klzns:
Yep, thats a breaking change

PR #139, opened 2019-05-03
Feature/janus client with account

guilhermebruzzi:

Add accountName option to JanusClient constructor

Allow to route to different account than in the request domain to a given request

Also added prettier

Also fixed tslint issues

I believe I can generate a beta tag to test it (open to suggestions)

#### Types of changes

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,8 @@
+{
+  "semi": false,
+  "singleQuote": true,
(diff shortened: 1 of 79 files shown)


firstdoit:
You can achieve the same by overwriting the `account` property in `context`:

```super({...context, account: 'banana'}, options)```

PR #142, opened 2019-05-06
Disable cache for mygocommerce.com

brunoabreu:
Disable cache for mygocommerce.com

GoCommerce admin requests should not be cached, otherwise the users will not see theirs changes happening.

#### Types of changes


victorges:
👀

PR #164, opened 2019-05-29
Allow environment selection, e.g beta

Jeymisson:
Allow environment selection, e.g beta, when cookie `vtex-commerce-env` is set

#### Types of changes

### DIFF
--- a/src/HttpClient/JanusClient.ts
+++ b/src/HttpClient/JanusClient.ts
@@ -20,6 +20,17 @@ export class JanusClient extends ExternalClient {
   constructor(context: IOContext, options?: InstanceOptions) {
     const {account} = context
 
+    if (context.janusEnv) {
+      options = options || {}
+      options = {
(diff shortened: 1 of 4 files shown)


src/HttpClient/JanusClient.ts
```diff
@@ -20,6 +20,17 @@ export class JanusClient extends ExternalClient {
...
+        ...options,
+        headers: {
+          ...options.headers,
+          'cookie': context.janusEnv,
+        },
+      }
+    }
+
     super(
       'http://portal.vtexcommercestable.com.br',
```
firstdoit: Following the `colossus-legacy-proxy` implementation, we should change host to `vtexcommercebeta` instead of passing a cookie. https://github.com/vtex/colossus-legacy-proxy/blob/master/node/index.ts#L47
Don't forget to add the outbound access policy in the consumer service.

PR #166, opened 2019-06-05
Add housekeeper client to node-vtex-api

vcalasans:
Implements the housekeeper client.

#### Types of changes


src/clients/Housekeeper.ts
```diff
@@ -0,0 +1,37 @@
+import { InfraClient, InstanceOptions } from '../HttpClient'
+import { HousekeeperStatesAndUpdates } from '../responses'
+import { IOContext } from '../service/typings'
+
+const createRoutes = ({account, workspace}: IOContext) => {
+  const routes = {
+    Apply: () => `v2/housekeeping/apply`,
+    Perform: () => `_housekeeping/perform`,
```
iagoaraujo: This endpoint is also `/v2`

src/responses.ts
```diff
@@ -110,3 +110,22 @@ export type AppBundlePublished = AppBundleResponse & {
 export type AppBundleLinked = AppBundleResponse & {
   bundleSize?: number,
 }
+
+export interface HousekeeperStatesAndUpdates {
+  state: {
+    infra: string[]
+    edition: string[]
+    runtimes: string[]
+    editionApps?: {
```
iagoaraujo: Replace `state` to be `updates` and `updates` to be `state`.
(`state` doesn't have the editionApps field. It's a field of updates)

Also I think you can separate the `state` and `updates` in a different interfaces (HousekeeperState and HousekeeperUpdates), and updates extends the state interface. Ex:

```
state {
    infra: string[]
    edition: string[]
    runtimes: string[]
    apps: Array<{id: string, source: string}>
  }

updates extends state {
    editionApps?: {
     ....
    }
}
```

PR #170, opened 2019-06-19
Include extensions on dev GraphQL responses

athoscouto:

To include extensions on dev GraphQL responses

This allows tracing information to be returned with the response.
Tracing information can be used to enhance profiling of GrahpQL resolvers.

#### Test instructions

1. Link any GraphQL app and will see an available service route for GraphQL (e.g. `http://styles-graphql.vtex.aws-us-east-1.vtex.io/storecomponents/athos/_v/graphql?__v=1.7.2`).
2. Using something like https://github.com/prisma/graphql-playground you can query that route. Since it is a private route, you'll need to add the proper authorization headers in the request.
3. With this changes, you'll receive extension data in the response.

On GraphQL Playground you'll be able to see tracing information like:

Screen Shot 2019-06-19 at 13 58 55

Instead of:

Screen Shot 2019-06-19 at 14 00 58

#### Types of changes

### DIFF
--- a/src/service/graphql/middlewares/response.ts
+++ b/src/service/graphql/middlewares/response.ts
@@ -6,6 +6,7 @@ import { cacheControl } from '../utils/cacheControl'
 
 export async function response (ctx: GraphQLServiceContext, next: () => Promise<void>) {
   const {responseInit, graphqlResponse} = ctx.graphql
+  const {production} = ctx.vtex
 
   const {
     maxAge = '',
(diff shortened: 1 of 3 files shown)


src/service/graphql/middlewares/response.ts
```diff
@@ -23,7 +24,10 @@ export async function response (ctx: GraphQLServiceContext, next: () => Promise<
     ctx.vary(SEGMENT_HEADER)
   }
 
-  ctx.body = pick(['data', 'errors'], graphqlResponse!)
+  const fields = production
+    ? ['data', 'errors']
+    : ['data', 'errors', 'extensions']
```
tlgimenes: Please move these arrays outside the function so we don't create then everytime

PR #177, opened 2019-07-02
Added head request and vbase-client getFileMetadata

marcosvcp:

As title says

#### Types of changes

### DIFF
--- a/src/HttpClient/HttpClient.ts
+++ b/src/HttpClient/HttpClient.ts
@@ -121,6 +121,11 @@ export class HttpClient {
     return this.request(patchConfig).then(response => response.data as T)
   }
 
+  public head = (url: string, config:RequestConfig = {}): Promise<IOResponse<void>> => {
+    const headConfig: RequestConfig = {...config, url, method: 'head'}
+    return this.request(headConfig)
+  }
+
   public delete = <T = void>(url: string, config?: RequestConfig): Promise<IOResponse<T>> => {
     const deleteConfig: RequestConfig = {...config, url, method: 'delete'}
     return this.request(deleteConfig)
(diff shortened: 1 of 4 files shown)


src/HttpClient/HttpClient.ts
```diff
@@ -121,6 +121,11 @@ export class HttpClient {
     return this.request(patchConfig).then(response => response.data as T)
   }
 
+  public head = (url: string, config:RequestConfig = {}): Promise<IOResponse<void>> => {
+    const headConfig: RequestConfig = {...config, url, method: 'head'}
```
tlgimenes: this is cacheable, right ?

src/clients/VBase.ts
```diff
@@ -72,6 +72,10 @@ export class VBase extends InfraClient {
     return this.saveContent(bucket, path, stream, {gzip, ttl})
   }
 
+  public getFileMetadata = (bucket:string, path:string) => {
```
tlgimenes: didn't know we supported this, cool !

src/HttpClient/HttpClient.ts
marcosvcp: Let's think ... HEAD is a method that has no body, it returns only headers that are literally metadata about the stored data and usually it MUST take less time than to make a GET obviously. When the stored data changes the HEAD should return me different metadata, and it is used as support for ETag for example, so I believe that it does not make sense to cache HEAD. Agrees that if we update an entity and the HEAD request returns the same thing because it was cached it becomes useless?

tlgimenes: hmmm. I have mixed feelings about this. I totally agree with you in the vbase case. However, think about an app that needs to `HEAD` another app. This head request will be performed without an `if-none-match` header, and the router won't be able to validade this request. As so, the request will be routed to the destiny app, making this app receive an unecessary request

tlgimenes: The only problem I see with cache is that I think our middlewares are not ready for caching `HEAD` requests

marcosvcp: As we commented on slack, it's not a problem not caching HEAD requests

PR #186, opened 2019-07-10
Send logs to stdout

firstdoit:


src/service/logger/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  Error = 'error',
+}
+
+export class Logger {
+  private account: string
+  private workspace: string
+  private operationId: string
+  private requestId: string
+
+  constructor(ctx: Partial<IOContext>) {
```
tlgimenes: should be a Pick in here, right ?

PR #188, opened 2019-07-19
Use GET with body for GraphQL queries

vcalasans:
As the name says.

### DIFF
--- a/src/HttpClient/IOClient.ts
+++ b/src/HttpClient/IOClient.ts
@@ -9,7 +9,7 @@ export type IOClientConstructor = new (context: IOContext, options?: InstanceOpt
  * A client that can be instantiated by the Service runtime layer.
  */
 export class IOClient {
-  protected http: HttpClient
+  public http: HttpClient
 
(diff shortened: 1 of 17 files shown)


src/HttpClient/IOClient.ts
```diff
@@ -9,7 +9,7 @@ export type IOClientConstructor = new (context: IOContext, options?: InstanceOpt
  * A client that can be instantiated by the Service runtime layer.
  */
 export class IOClient {
-  protected http: HttpClient
+  public http: HttpClient
```
tlgimenes: why ?

src/service/graphql/middlewares/query.ts
```diff
@@ -1,9 +1,14 @@
-import { json } from 'co-body'
+import { default as bodyParse } from 'co-body'
 import { compose, partialRight, prop } from 'ramda'
 import { parse, Url } from 'url'
 import { GraphQLServiceContext } from '../typings'
 
+const BODY_HASH = 'bodyHash'
```
tlgimenes: why not using a more obscure variable like `__graphqlBodyHash`

PR #190, opened 2019-07-23
Add x-vtex-product to open segments for specific produt

Jeymisson:
It passes the querystrings `session_path` and `session_host`, which are used to separate products when we open message's segments.

#### Types of changes

### DIFF
--- a/src/clients/Segment.ts
+++ b/src/clients/Segment.ts
@@ -78,13 +78,13 @@ export class Segment extends JanusClient {
       forceMaxAge: SEGMENT_MAX_AGE_S,
       headers: {
         'Content-Type': 'application/json',
-        [PRODUCT_HEADER]: product,
+        [PRODUCT_HEADER]: product || '',
       },
       inflightKey: inflightUrlWithQuery,
       metric: 'segment-get',
       params: {
         ...sanitizeParams(query),
-        session_path: product,
+        session_path: product || '',
       },
     }))
   }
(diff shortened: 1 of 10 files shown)


src/clients/Segment.ts
```diff
@@ -71,15 +72,19 @@ export class Segment extends JanusClient {
   }
 
   private rawSegment = (token?: string | null, query?: Record<string, string>) => {
+    const { product } = this.context
+
     return this.http.getRaw<SegmentData>(routes.segments(token), ({
       forceMaxAge: SEGMENT_MAX_AGE_S,
       headers: {
         'Content-Type': 'application/json',
+        [PRODUCT_HEADER]: product,
```
brunoabreu: I think axios explodes if you set a header with no value. It's safer to add this header only if `product` is there.

src/clients/Segment.ts
```diff
@@ -71,15 +72,19 @@ export class Segment extends JanusClient {
...
       },
       inflightKey: inflightUrlWithQuery,
       metric: 'segment-get',
       params: {
         ...sanitizeParams(query),
+        session_path: product,
```
brunoabreu: Same for `session_path`. It won't explode, but I'm not sure it won't send `?session_path=undefined` 🤔

Jeymisson: Sure! We expect to always have product, but I will protect it

PR #108, opened 2019-03-31
Deprecates listing files api from VBase client

tlgimenes:
#### What is the purpose of this pull request?
Deprecates the `/files` API used in Vbase client due to changes in API

#### What problem is this solving?
Vbase is currently deprecating files listing. This PR solves this problem by removing the endpoint used by Vbase client. However, the methods were not removed from Vbase client. When these methods are called, an exception is thrown with a deprecation message.

#### Types of changes

* [ ] Bug fix (a non-breaking change which fixes an issue)
* [ ] New feature (a non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Requires change to documentation, which has been updated accordingly.

PR #196, opened 2019-07-25
Remove `list` from vbase client

vcalasans:
Since it is no longer supported.


tlgimenes:
Also, there is #108 doing the same thing

PR #197, opened 2019-07-26
Fix caching

thiagomurakami:

Delete headers and set vary for graphql routes according to scope.

Public routes that should be cached.

???

#### Types of changes


src/service/http/middlewares/vary.ts
```diff
@@ -1,22 +1,81 @@
...
+function isGraphqlContext <T extends IOClients, U, V> (ctx: ServiceContext<T,U,V>): ctx is ServiceContext<T,U,V> & Pick<GraphQLContext, 'graphql'> {
+  if ('graphql' in ctx) {
+    return true
+  }
+  return false
+}
+
+function setVaryIfExists<T extends IOClients, U, V> (ctx: ServiceContext<T, U, V>, varyHeaders: string[]) {
+  varyHeaders.forEach((varyHeader) => {
+    if (ctx.get(varyHeader)) {
```
brunoabreu: We should always add `vary`, even if the header is not there.

src/service/http/middlewares/vary.ts
```diff
@@ -1,22 +1,81 @@
...
+      ctx.vary(SEGMENT_HEADER)
+    }
+    if (ctx.get(SESSION_HEADER)) {
+      ctx.vary(SESSION_HEADER)
+    }
   }
 
   await next()
+
+  if (isGraphql && isGraphqlContext(ctx)) {
```
brunoabreu: I'm not happy with this middleware being aware of graphql, but I can't think a better way to do it 😞

src/service/http/middlewares/cdnNormalizer.ts
```diff
@@ -0,0 +1,41 @@
...
+    forbidden: [SESSION_HEADER],
+    path: '/_v/segment.',
+    vary: [SEGMENT_HEADER],
+  },
+]
+
+
+export async function cdnNormalizer <T extends IOClients, U, V> (ctx: ServiceContext<T, U, V>, next: () => Promise<any>) {
+  const { path } = ctx
+  const strategy = cachingStrategies.find(cachingStrategy => path.indexOf(cachingStrategy.path) === 0)
```
thiagomurakami: https://jsperf.com/publicregexstartswith

src/service/graphql/middlewares/response.ts
```diff
@@ -23,10 +21,6 @@ export async function response (ctx: GraphQLServiceContext, next: () => Promise<
     'Cache-Control': cacheControlHeader,
   })
 
-  if (segment) {
```
brunoabreu: 🤔

PR #202, opened 2019-08-02
Replace tslint with eslint

lucasecdb:
Replace the deprecated linter `tslint` with `eslint` and the default configurations from `eslint-config-vtex`.

TSLint will be deprecated soon and should be replaced by ESLint.

#### Types of changes

* [x] Requires change to documentation, which has been updated accordingly.

### DIFF
--- a/src/HttpClient/middlewares/inflight.ts
+++ b/src/HttpClient/middlewares/inflight.ts
@@ -1,4 +1,3 @@
-import { createHash } from 'crypto'
 import { stringify } from 'qs'
 import { InflightKeyGenerator, MiddlewareContext, RequestConfig } from '../typings'
 
@@ -38,20 +37,18 @@ export const singleFlightMiddleware = async (ctx: MiddlewareContext, next: () =>
     ctx.response = memoized.response
     return
   } else {
-    const promise = new Promise<Inflight>(async (resolve, reject) => {
-      try {
-        await next()
-        resolve({
-          cacheHit: ctx.cacheHit!,
-          response: ctx.response!,
+    const promise = new Promise<Inflight>((resolve, reject) => {
+      next()
+        .then(() => {
+          resolve({
+            cacheHit: ctx.cacheHit!,
+            response: ctx.response!,
+          })
+        }).catch((err) => {
+          reject(err)
+        }).finally (() => {
+          inflight.delete(key)
         })
-      }
-      catch (err) {
-        reject(err)
-      }
-      finally {
-        inflight.delete(key)
-      }
     })
     inflight.set(key, promise)
     await promise
--- a/src/HttpClient/middlewares/memoization.ts
+++ b/src/HttpClient/middlewares/memoization.ts
@@ -29,17 +29,13 @@ export const memoizationMiddleware = ({memoizedCache}: MemoizationOptions) => {
       ctx.response = memoized.response
       return
     } else {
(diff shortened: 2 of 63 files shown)


src/HttpClient/middlewares/inflight.ts
```diff
@@ -38,20 +37,18 @@ export const singleFlightMiddleware = async (ctx: MiddlewareContext, next: () =>
...
     return
   } else {
-    const promise = new Promise<Inflight>(async (resolve, reject) => {
-      try {
-        await next()
-        resolve({
-          cacheHit: ctx.cacheHit!,
-          response: ctx.response!,
+    const promise = new Promise<Inflight>((resolve, reject) => {
+      next()
```
lucasecdb: I removed the async function here because of the [`no-async-promise-executor`](https://eslint.org/docs/rules/no-async-promise-executor) rule, it may not accomplish the same as it did before

src/HttpClient/middlewares/memoization.ts
```diff
@@ -29,17 +29,13 @@ export const memoizationMiddleware = ({memoizedCache}: MemoizationOptions) => {
       ctx.response = memoized.response
       return
     } else {
-      const promise = new Promise<Memoized>(async (resolve, reject) => {
-        try {
-          await next()
+      const promise = new Promise<Memoized>((resolve, reject) => {
+        next().then(() => {
```
lucasecdb: same here

luciannojunior:
Update CHANGELOG.md

src/HttpClient/middlewares/inflight.ts
```diff
@@ -38,20 +37,18 @@ export const singleFlightMiddleware = async (ctx: MiddlewareContext, next: () =>
...
+        .then(() => {
+          resolve({
```
tlgimenes: what if resolve throws ?

src/HttpClient/middlewares/metrics.ts
```diff
@@ -40,7 +40,7 @@ export const metricsMiddleware = ({metrics, serverTiming, name}: MetricsOpts) =>
   }))
   return async (ctx: MiddlewareContext, next: () => Promise<void>) => {
     const start = process.hrtime()
-    let status: string = 'unknown'
+    let status = 'unknown'
```
tlgimenes: it has a diferent meaning, no ?

src/clients/Apps.ts
```diff
@@ -217,12 +216,14 @@ export class Apps extends InfraClient {
     const locator = parseAppId(app)
     const linked = !!locator.build
     const inflightKey = inflightURL
-    return this.http.get<T>(this.routes.File(locator, path), {
+    const config: IgnoreNotFoundRequestConfig = {
+      // @ts-ignore
```
tlgimenes: why ignore in here ?

src/HttpClient/middlewares/metrics.ts
lucasecdb: what do you mean? typescript will still infer the `status` type as being a `string`, right?

src/clients/Apps.ts
lucasecdb: the `cacheable` property doesn't exist on the type `IgnoreNotFoundRequestConfig`, so I need to ignore the error.

we didn't need this before because it was using the object literal assertion, so typescript assumed everything was ok and didn't typecheck the properties

PR #205, opened 2019-08-08
Deal better with events

R-Araripe:
Improves event handler to support clients and piping of middlewares, as well as normalizes the context ( for example, includes vtex property of type IOContext in events context, as it is the format for http routes context).

#### Types of changes


src/service/events/index.ts
```diff
@@ -0,0 +1,20 @@
...
+import { timings } from './middlewares/timings'
+
+export const createEventHandler = <ClientsT extends IOClients, StateT>(
+  Clients: ClientsImplementation<ClientsT>,
+  options: Record<string, InstanceOptions>
+) => {
+  return (handler: EventHandler<ClientsT, StateT> | Array<EventHandler<ClientsT, StateT>>) => {
+    const middlewares = Array.isArray(handler) ? handler : [handler]
+    const pipeline = [contextNormalizer, clients(Clients, options), timings, error, ...middlewares]
+    return composeForEvents(pipeline)
```
vcalasans: Can't we use the `compose` function from `../utils/compose` here?

src/service/events/middlewares/contextNormalizer.ts
```diff
@@ -0,0 +1,9 @@
+import { pick } from 'ramda'
+
+export async function contextNormalizer (ctx: any, next: () => Promise<any>) {
+
+  ctx.vtex = pick(['account', 'workspace' , 'authToken' , 'region' , 'production' , 'userAgent' , 'segmentToken' , 'sessionToken' , 'requestId' , 'operationId' , 'product'], ctx)
```
vcalasans: Nit.: Spacing between word and `,`

src/service/utils/compose.ts
```diff
@@ -1,8 +1,11 @@
...
 import { IOClients } from '../../clients/IOClients'
-import { timer } from '../../utils/time'
-import { RouteHandler } from '../typings'
+import { timer, timerForEvents } from '../../utils/time'
+import { EventHandler, RouteHandler } from '../typings'
 
 export const compose = <ClientsT extends IOClients, StateT, CustomT>(middlewares: Array<RouteHandler<ClientsT, StateT, CustomT>>) =>
   koaCompose(middlewares.map(timer))
+
+export const composeForEvents = <ClientsT extends IOClients, StateT>(middlewares: Array<EventHandler<ClientsT, StateT>>) =>
```
vcalasans: As commented above, do we really need this function?

src/service/events/index.ts
R-Araripe: I don't think so, because the event handler type is different than for routes..

R-Araripe: Well we could use `koaCompose` from 'koa-compose' directly instead?

PR #213, opened 2019-08-16
Fix translate arg in MessagesGraphQL

R-Araripe:
At translate args in MessageGraphQL client, the input messages type should include the property provider for a list of messages.

#### Types of changes

### DIFF
--- a/src/clients/MessagesGraphQL.ts
+++ b/src/clients/MessagesGraphQL.ts
@@ -6,12 +6,17 @@ import { IOMessage } from '../utils/message'
 
 type IOMessageInput = Pick<IOMessage, 'id' | 'content' | 'description' | 'behavior'>
 
+interface IOMessagesInput {
+  provider: string,
+  messages: IOMessageInput[],
+}
+
 export interface IOMessageSaveInput extends IOMessageInput {
   content: string
 }
 
 export interface Translate {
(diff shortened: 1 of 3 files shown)


src/clients/MessagesGraphQL.ts
```diff
@@ -6,12 +6,17 @@ import { IOMessage } from '../utils/message'
 
 type IOMessageInput = Pick<IOMessage, 'id' | 'content' | 'description' | 'behavior'>
 
+interface IOMessagesInput {
```
jgfidelis: you now have a type and interface with the same name. Maybe give this another name?

tlgimenes:
```suggestion
interface MessagesInput {
```

src/clients/MessagesGraphQL.ts
```diff
@@ -6,12 +6,17 @@ import { IOMessage } from '../utils/message'
...
+  messages: IOMessageInput[],
+}
+
 export interface IOMessageSaveInput extends IOMessageInput {
   content: string
 }
 
 export interface Translate {
-  messages: IOMessageInput[]
+  messages: IOMessagesInput[]
```
tlgimenes:
```suggestion
  messages: MessagesInput[]
```

PR #223, opened 2019-08-29
Feature/make alternative translate directive calling translate messages by

R-Araripe:
Creates an alternative translatable directive, @translatable2. Besides behavior, It can take a context as an argument. It calls translate resolver in Messages (via MessagesGraphQL client) that uses the concept of "message as its own id".

#### Types of changes


src/clients/MessagesGraphQL.ts
```diff
@@ -49,6 +71,17 @@ export class MessagesGraphQL extends AppGraphQLClient {
...
 
+  public translateV2 = (args: TranslateInputV2) => this.graphql.query<TranslatedV2, { args: TranslateInputV2 }>({
+      query: `
+      query Translate($args: TranslateArgs!) {
+        translate(args: $args)
+      }
+      `,
+      variables: { args },
+    }, {
+      metric: 'messages-translate',
```
tlgimenes:
```suggestion
      metric: 'messages-translate-v2',
```

PR #225, opened 2019-09-02
Add environment arg to the JanusClient api.

salesfelipe:

- ditto.

- With this, it will be possible to point to beta even without the cookie `vtex-commerce-env`.

#### Types of changes

### DIFF
--- a/src/HttpClient/JanusClient.ts
+++ b/src/HttpClient/JanusClient.ts
@@ -3,6 +3,8 @@ import { IOContext } from '../service/typings'
 import { ExternalClient } from './ExternalClient'
 import { InstanceOptions } from './typings'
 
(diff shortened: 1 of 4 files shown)


src/HttpClient/JanusClient.ts
```diff
@@ -3,6 +3,11 @@ import { IOContext } from '../service/typings'
 import { ExternalClient } from './ExternalClient'
 import { InstanceOptions } from './typings'
 
+enum Environment {
```
tlgimenes: why not using string type literal ?
```suggestion
type Environment = 'beta' | 'stable' 
```

salesfelipe: I do prefer enum. For me is more readable and easier to write.

PR #235, opened 2019-09-13
Feature/cancellation token

NataliaGodot:

Create a cancellation token that terminates the processing of requests that have been canceled by the user, as well as the connections created by these requests.

Prevents unnecessary use of resources that is caused by processing requests that have already been canceled by the user.

#### Types of changes


src/HttpClient/middlewares/cancelationToken.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const cancelationToken = (cancelation?: Cancelation) => async (ctx: MiddlewareContext, next: () => Promise<void>) => {
+  const { config: { method } } = ctx
+
+  if (!cancelation) {
+    await next()
+    return
+  }
+  if (method && method.toUpperCase() !== 'GET') {
+    cancelation.cancelable = false
```
brunoabreu: @tlgimenes
I don't like it because we already ensured the request is safe (we only create `ctx.vtex.cancellation` for **GET**). A server saving data somewhere with **PUT** or **POST** while handling a **GET** request is something unusual and I don't think we should support this weird "GET transactions consistency".

If someone needs to avoid request process cancelation in a **GET** request, can do it explicitly in a middleware that deletes `ctx.vtex.cancelation` under certain circumstances.

PR #239, opened 2019-09-25
Add Mine Wins Conflict Resolver logic to VBase client

Jeymisson:
Add a solution strategy to keep updating selected workspaces configs with master’s configs whenever we can, but in case of a key: value conflict, we keep workspace’s (mine) values to maintain the work in progress of the workspace.

Loss of determined configs in workspaces when content is changed in master because our rebase policy always selects master's configs.

No, we have unit tests! yay

#### Types of changes


src/clients/VBase.ts
```diff
@@ -58,10 +59,19 @@ export class VBase extends InfraClient {
...
   }
 
-  public getJSON = <T>(bucket: string, path: string, nullIfNotFound?: boolean) => {
+  public getJSON = <T>(bucket: string, path: string, nullIfNotFound?: boolean, conflictsResolver?: ConflictsResolver) => {
+    const headers = conflictsResolver? {'X-Vtex-Detect-Conflicts': true}: {}
     const inflightKey = inflightURL
     const metric = 'vbase-get-json'
-    return this.http.get<T>(routes.File(bucket, path), {nullIfNotFound, metric, inflightKey} as IgnoreNotFoundRequestConfig)
+    return this.http.get<T>(routes.File(bucket, path), {nullIfNotFound, metric, inflightKey, headers} as IgnoreNotFoundRequestConfig).then((data)=>{
+      return data
```
brunoabreu: Why `.then((data) => { return data })`? It does nothing.

src/clients/VBase.ts
```diff
@@ -58,10 +59,19 @@ export class VBase extends InfraClient {
...
+    }).catch(error =>{
+      const {response} = error
+      if(response.status === 409 && conflictsResolver){
+        return conflictsResolver.resolve()
+      }
+      throw Error(error)
```
brunoabreu: Just throw the original error: `throw error`

PR #252, opened 2019-10-17
Assests client

vwraposo:
Creates new client called `Assets`.

This client is supposed to be used instead of `settings-server` beacause it gets the app's assets according to the vendor and not the account (if the app is not linked) this allows for better caching and unearths the true potential of passing the depTrees from `render-server` to the other services.

#### Types of changes

### DIFF
--- a/src/clients/Apps.ts
+++ b/src/clients/Apps.ts
@@ -85,7 +85,7 @@ export class Apps extends InfraClient {
   constructor(context: IOContext, options?: InstanceOptions) {
     super('apps', context, options, true)
     this.diskCache = options && options.diskCache
-    this.memoryCache = options && options.diskCache
+    this.memoryCache = options && options.memoryCache
     this._routes = createRoutes(context)
   }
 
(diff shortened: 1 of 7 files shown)


src/clients/Apps.ts
```diff
@@ -85,7 +85,7 @@ export class Apps extends InfraClient {
   constructor(context: IOContext, options?: InstanceOptions) {
     super('apps', context, options, true)
     this.diskCache = options && options.diskCache
-    this.memoryCache = options && options.diskCache
+    this.memoryCache = options && options.memoryCache
```
Jeymisson: Holy cow!

src/clients/Assets.ts
```diff
@@ -0,0 +1,122 @@
...
+  public getAppJSONByVendor = <T extends object | null>(app: string, path: string, nullIfNotFound?: boolean) => {
+    const locator = parseAppId(app)
+    const vendor = locator.name.split('.')[0]
+    const linked = !!locator.build
+    const inflightKey = inflightURL
+    return linked? this.getAppJSON(app, path, nullIfNotFound):
+      this.http.get<T>(this.route(vendor, locator, path), {
+        cacheable: CacheType.Any,
+        inflightKey,
+        metric: 'registry-get-assets-json-by-vendor',
```
Jeymisson: Maybe we want this same logic here `metric: linked ? 'apps-get-assets-json' : 'registry-get-assets-json-by-vendor'`

src/clients/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './Apps'
+export * from './Assets'
```
Jeymisson: Not sure if assets is the best name, but I have no suggestions other than `settings` that's already in use.

src/clients/Assets.ts
```diff
@@ -0,0 +1,122 @@
...
+  constructor(context: IOContext, options?: InstanceOptions) {
+    super('apps', context, options, true)
+    this.route = this.fileRoute(this.context.workspace)
+  }
+
+  public async getSettings (dependencies: AppMetaInfo[], appAtMajor: string, params: AssetsParams = {}) {
+    const filtered = this.getFilteredDependencies(appAtMajor, dependencies)
+    const {pick, files} = params
+
+    return await Promise.all(filtered.map(dependency => {
```
tlgimenes: why do you need this await in here ?

src/clients/Assets.ts
```diff
@@ -0,0 +1,122 @@
...
+    const result = !isEmpty(pickArray) ? ramdaPick(pickArray, buildJson) : buildJson
+
+    result.declarer = app.id
+    return result
+  }
+
+  public async getSettingsFromFilesForApp(app: AppMetaInfo, files: string | string[] = []): Promise<Record<string, any>> {
+    // If there's no support for build.json, then fetch individual files and zip them into an {[file]: content} object.
+    const filesArray = Array.isArray(files) ? files : [files]
+    const fetched = await Promise.all(map((file) => this.getJSON(app.id, file, true), filesArray as string[]))
```
tlgimenes: why using ramda's `map` in here ? Why forcing `filesArray` as `string[]` in here ? doesn't the compiler gets the typings right in here ?

PR #255, opened 2019-10-21
Handle nullable arrays passed to MineWinsConflictsResolver

Jeymisson:
#### Types of changes

### DIFF
--- a/src/utils/MineWinsConflictsResolver.ts
+++ b/src/utils/MineWinsConflictsResolver.ts
@@ -4,7 +4,7 @@ import { eqProps, equals } from 'ramda'
 import { isArray, isObject } from 'util'
 import { ConflictsResolver, VBase, VBaseConflict, VBaseConflictData } from '../clients/VBase'
 
-type Configuration = Record<string, object | object[]> | object[]
+type Configuration = Record<string, object | object[] | null> | object[]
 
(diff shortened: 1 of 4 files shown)


src/utils/MineWinsConflictsResolver.ts
```diff
@@ -115,37 +115,48 @@ export class MineWinsConflictsResolver implements ConflictsResolver {
...
+            mine.splice(foundInMine, 1)
+          }
         }
-      }
-    })
+      })
+    }
   }
 
   private isObjectInArray(obj: object, array: object[]) {
```
jgfidelis: I think its better to change the types to reflect that it may be nullable, like `array: object[] | null`.

Otherwise months from today, someone will see the type and think that this check in line 132 is irrelevant.

src/utils/MineWinsConflictsResolver.ts
```diff
@@ -115,37 +115,48 @@ export class MineWinsConflictsResolver implements ConflictsResolver {
   }
 
   private removeMasterDeletedElements(base: object[], master: object[], mine: object[]) {
```
jgfidelis: base is not of type `object[]`

src/utils/MineWinsConflictsResolver.ts
Jeymisson: Fair enough

PR #282, opened 2019-11-28
Update publish endpoint

iagoaraujo:
Deprecate the current publish API and starts to use the new one: https://github.com/vtex/apps/blob/master/api/v2/registryController.go#L34

#### Types of changes

### DIFF
--- a/src/clients/Registry.ts
+++ b/src/clients/Registry.ts
@@ -20,7 +20,7 @@ const routes = {
   AppFile: (app: string, version: string, path: string) => `${routes.AppFiles(app, version)}/${path}`,
   AppFiles: (app: string, version: string) => `${routes.AppVersion(app, version)}/files`,
   AppVersion: (app: string, version: string) => `${routes.App(app)}/${version}`,
-  Publish: '/v2/registry',
+  Publish: '/v2/registry/rc',
   Registry: '/registry',
   ResolveDependenciesWithManifest: '/v2/registry/_resolve',
 }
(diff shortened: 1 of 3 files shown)


src/clients/Registry.ts
```diff
@@ -77,7 +77,17 @@ export class Registry extends InfraClient {
 
   public deprecateApp = (app: string, version: string) => {
     const metric = 'registry-deprecate'
-    return this.http.patch(routes.AppVersion(app, version), {deprecated: true}, {metric})
+    return this.http.patch(routes.AppVersion(app, version), {patchState: "deprecate"}, {metric})
+  }
+
+  public unDeprecateApp = (app: string, version: string) => {
```
tiagonapoli: nit: I think this could be `undeprecateApp`

src/clients/Registry.ts
```diff
@@ -20,7 +20,7 @@ const routes = {
   AppFile: (app: string, version: string, path: string) => `${routes.AppFiles(app, version)}/${path}`,
   AppFiles: (app: string, version: string) => `${routes.AppVersion(app, version)}/files`,
   AppVersion: (app: string, version: string) => `${routes.App(app)}/${version}`,
-  Publish: '/v2/registry',
+  Publish: '/v2/registry/rc',
```
tergol1: With this change we'll immediately start to release apps as release candidates, right?

tergol1: Just what I thought: create a new function which uses the new API and let the change in publish behavior to the one that comes in Builder-Hub, which will migrate from one API to another.

iagoaraujo: Ok.
But you don't need to worry, only builder-hub can publish apps and the node-vtex-api will be updated only when we release the runtime-node with this version of node-vtex-api.

tergol1: Ahh Ok! I understood!

PR #287, opened 2019-12-09
Meta infos stale while revalidating

vwraposo:
Makes `Apps.getAppsMetaInfos` return stale and then validate and update the cache

This will allow pages to be rendered without Apps

#### Types of changes

### DIFF
--- a/src/clients/Apps.ts
+++ b/src/clients/Apps.ts
@@ -306,9 +306,7 @@ export class Apps extends InfraClient {
     const key = getMetaInfoKey(account)
 
     const cachedResponse: {appsMetaInfo: AppMetaInfo[], headers: any} | undefined = this.diskCache && await this.diskCache.get(key)
-    if (!this.context.recorder) {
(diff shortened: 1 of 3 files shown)


src/clients/Apps.ts
```diff
@@ -299,25 +299,39 @@ export class Apps extends InfraClient {
...
+        return response
+      })
+
+    const useCachedResponse = cachedResponse && production && staleWhileRevalidate
+    const appsMetaInfo: AppMetaInfo[] = useCachedResponse
+      ? cachedResponse!.appsMetaInfo
+      : await metaInfoPromise.then(response => response.data.apps)
+
+    if (filter) {
+      return ramdaFilter(appMeta => !!ramdaPath(['_resolvedDependencies', filter], appMeta), appsMetaInfo)
```
NataliaGodot: NIT
return appsMetaInfo.filter(appMeta => !!ramdaPath(['_resolvedDependencies', filter], appMeta))

PR #306, opened 2020-01-13
Recorder Performance

tlgimenes:
This PR should help us tackle the recorder performance issue in `vtex.messages` and in all other apps that are heavily dependent on `x-vtex-meta` headers
The performance issue emerged when the recorder was used multiple times. Since each insertion of a new `x-vtex-meta` header parsed the whole header value (with splits and trims), each insertion was costly. The proposed fix makes the insertion of a new value cheap by aggregating  each value of a insert in a `Set` structure. After all headers are inserted in the `Set`, we make a `flush` operation, where all the values are parsed and formatted once.

This PR decreased up to 50ms of blocking event loop time in `vtex.messages` and I hope this should lower the spikes seen in `vtex.messages` and in `vtex.pages-graphql`

Since `vtex.messages` uses the recorder, this PR doesn't enable the new recorder behavior by default. In stead, it only exports the new recorder interface so we can use it in messages as a preparation for activating the FAST_RECORDER feature flag

#### Types of changes


src/service/worker/runtime/utils/recorder.ts
```diff
@@ -16,7 +19,51 @@ const appendResponseHeader = (ctx: Context, responseHeaders: any, targetHeader:
   }
 }
 
-export const createRecorder = (ctx: Context) => (headers: any) => {
+export const createSlowRecorder = (ctx: Context): SlowRecorder => (headers: Record<string, string>) => {
   appendResponseHeader(ctx, headers, 'x-vtex-meta')
   appendResponseHeader(ctx, headers, 'x-vtex-meta-bucket')
```
tiagonapoli: Could `META_HEADER` and `META_HEADER_BUCKED` be used here?

src/service/worker/runtime/utils/recorder.ts
```diff
@@ -16,7 +19,51 @@ const appendResponseHeader = (ctx: Context, responseHeaders: any, targetHeader:
...
 }
+
+export const isSlowRecorder = (x: any): x is SlowRecorder => typeof x === 'function' && !x._record
```
tiagonapoli: Nice, I was wondering if `instanceof x === 'Recorder'` could be used here but it's actually slower than `typeof`

tlgimenes: 👍

src/service/worker/runtime/utils/recorder.ts
```diff
@@ -16,7 +19,51 @@ const appendResponseHeader = (ctx: Context, responseHeaders: any, targetHeader:
...
+    this._record = HEADERS.reduce(
+      (acc, headerName) => {
+        acc[headerName] = new Set<string>()
+        return acc
+      },
+      {} as Record<string, Set<string>>
+    )
+  }
+
+  public clear = () => HEADERS.forEach(headerName => this._record[headerName].clear())
```
tiagonapoli: If class functions are not used within other `this` environments, using:
```
public clear() {
 ...
}
```
Is slightly faster - mainly if the class is often instantiated.

The source from this info I got from https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1 - is 2 years old, so maybe this scenario is already better, but the performance drawbacks explained there make sense. What do you think?

src/service/worker/runtime/utils/recorder.ts
tlgimenes: They could !

src/service/worker/runtime/utils/recorder.ts
tlgimenes: Nice, I didn't know about these performance issues

src/service/worker/runtime/utils/recorder.ts
```diff
@@ -1,8 +1,11 @@
 import { Context } from 'koa'
 import { trim, uniqWith } from 'ramda'
 
+import { META_HEADER, META_HEADER_BUCKET } from './../../../../constants'
 import { toArray } from './toArray'
 
+export type SlowRecorder = (headers: Record<string, string>) => void
+
 // uniqWith is way faster than ramda's uniq
 const uniqStr = uniqWith((a: string, b: string) => a === b)
```
jgfidelis: this can be `const uniqStr =  uniqBy(val => val)`

src/service/worker/runtime/utils/recorder.ts
```diff
@@ -16,7 +19,57 @@ const appendResponseHeader = (ctx: Context, responseHeaders: any, targetHeader:
...
+
+  public flush (ctx: Context) {
+    HEADERS.forEach(
+      headerName => {
+        const currentValue = ctx.response.get(headerName) as string | string[]
+        const parsedCurrentValue = typeof currentValue === 'string'
+          ? currentValue.split(',')
+          : currentValue
+        const toAppend = Array.from(this._record[headerName])
+        const deduped = uniqStr([...parsedCurrentValue, ...toAppend].filter(x => !!x))
```
tiagonapoli: Maybe the values on `parsedCurrentValue` would need to be trimmed?

src/service/worker/runtime/utils/recorder.ts
```diff
@@ -16,7 +19,57 @@ const appendResponseHeader = (ctx: Context, responseHeaders: any, targetHeader:
...
+        if (deduped.length > 0) {
+          ctx.set(headerName, deduped)
```
tiagonapoli: Does `ctx.set` work with arrays? (I couldn't find docs :disappointed:)

tiagonapoli: Just found

src/service/worker/runtime/utils/recorder.ts
tlgimenes: According to the ramda doc, https://ramdajs.com/docs/#uniqBy, uniqBy uses R.equals for comparing elements. R.equals cause a HUGE performance degradation

src/service/worker/runtime/utils/recorder.ts
tlgimenes: Thanks ! I think it should

PR #307, opened 2020-01-14
implement retries with custom parameters for increasing timeout and d…

jgfidelis:
…elay

#### Types of changes

### DIFF
--- a/src/HttpClient/middlewares/request.ts
+++ b/src/HttpClient/middlewares/request.ts
@@ -87,15 +87,15 @@ function fixConfig(axiosInstance: AxiosInstance, config: RequestConfig) {
   }
 }
 
-const customExponentialDelay = (initialBackoffDelay: number, exponentialBackoffCoefficient: number, retryNumber: number) => {
-  const delay = initialBackoffDelay * Math.pow(exponentialBackoffCoefficient, retryNumber)
+const customExponentialDelay = (backoffDelayConstant: number, exponentialBackoffCoefficient: number, retryNumber: number) => {
+  const delay = backoffDelayConstant * Math.pow(exponentialBackoffCoefficient, retryNumber)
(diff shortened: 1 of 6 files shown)


src/HttpClient/middlewares/request.ts
```diff
@@ -19,23 +18,74 @@ const http = axios.create({
...
+
+  if (config.retries == null || config.retries === 0) {
+    return Promise.reject(error)
+  }
+  const { backoffDelayConstant = 100, exponentialBackoffCoefficient = 2, exponentialTimeoutCoefficient } = config
+  const retryCount = config.retryCount || 0
+  const shouldRetry = isAbortedOrNetworkErrorOrRouterTimeout(error) && retryCount < config.retries
+  if (shouldRetry) {
+    config.retryCount = retryCount + 1
+    const delay = exponentialDelay(backoffDelayConstant, exponentialBackoffCoefficient, config.retryCount)
```
brunoabreu: I think the first one should be with `config.retryCount - 1` or just `retryCount` (before increment). Otherwise the first backoff will always be bigger than my `backoffDelayConstant`.

If I set `backoffDelayConstant` to 1s and `exponentialBackoffCoefficient` to 2, the first backoff will be ~2s.. that's weird.

Maybe the best way to fix it is changing the `exponentialDelay` function to use `retryCount - 1` as exponent.

jgfidelis: :peposad:
I thought about this when implementing and that is exactly why I changed the name from `initialBackoffDelay` to `backoffDelayConstant`. Its just a way you can set a custom variation to the backoff formula. The formula will always be `delay = constant * coefficient^retryCount` . You can tweak the constant and coefficient as you wish and get the desired values. Do you still prefer to have the first retry not using any formula but using the constant value? If yes, I am fine with that.

PR #313, opened 2020-01-22
Feature/service configuration context

dhasuda:
Add the settings middleware. With this modifications, services will receive their configurations from other apps in the request context.

Better interface for getting configurations.

Create a service with `workspace` or `userAndWorkspace` settings type. After that, create an app with configurations to be exported to this service. Link them all and see if the request context comes with the configurations.

In more detail, in your `service` you must have:

1. A route with settings type setted as `workspace` or `userAndWorkspace`. For that, add something like this to your `node/service.json`:
```
"status": {
      "path": "/_v/status/:code",
      "public": true,
      "settingsType": "workspace"
    }
```

2. In the `manifest.json`, don't forget to add the `configuration` builder:
```
"builders": {
    "configuration": "0.x",
    ...
  },
...
```

3. Add the file `configuration/schema.json` with a schema for the configuration, for instance:
```
{
  "type": "object",
  "properties": {
    "id": { "type": "number" },
    "name": { "type": "string" }
  }
}
```

In your `app` that exports the configuration, you must have:
1. In the `manifest.json`, a declaration of the service as a builder:
```
"builders": {
    "vtex.<serviceName>": "0.x",
    ...
  },
```

2. The file `vtex.<serviceName>/configuration.json` must have a configuration that conforms to schema in the service:
```
{
  "name": "little foot",
  "id": 19
}
```

Link first the service, and then the app. Make the request and check in `ctx.vtex.settings` if there are the exported settings

#### Types of changes


src/service/worker/runtime/graphql/schema/schemaDirectives/Settings.ts
```diff
@@ -0,0 +1,41 @@
+import { defaultFieldResolver, GraphQLField } from 'graphql'
+import { SchemaDirectiveVisitor } from 'graphql-tools'
+import { Apps } from '../../../../../../clients/infra/Apps'
+import { Assets } from '../../../../../../clients/infra/Assets'
+import { getDependenciesSettings } from '../../../http/middlewares/settings'
+import { RouteSettingsType } from '../../../typings'
+
+const getSettings = async (settingsType: any, ctx: any) => {
+  const settings = settingsType as RouteSettingsType
```
vwraposo: Do you need to typecast? Can't you have it correctly in the function parameters?

src/service/worker/runtime/graphql/schema/schemaDirectives/Settings.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (settings !== 'workspace' && settings !== 'userAndWorkspace') { return ctx }
```
vwraposo: NIT: Isn't this the same as `settings === 'pure'`

src/service/worker/runtime/graphql/schema/schemaDirectives/Settings.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const { clients: { apps, assets } } = ctx
+  const dependenciesSettings = await getDependenciesSettings(apps as Apps, assets)
+  ctx = {
+    ...ctx,
+    vtex: {
+      settings: dependenciesSettings,
+      ...ctx.vtex,
+    },
+  }
```
vwraposo: I think it is better to do `ctx.vtex.settings = ctx.vtex.settings || dependenciesSettings`, to avoid too many spread operations

src/service/worker/runtime/graphql/schema/schemaDirectives/Settings.ts
```diff
@@ -0,0 +1,41 @@
...
+  return ctx
+}
+
+export class SettingsDirective extends SchemaDirectiveVisitor {
+  public visitFieldDefinition (field: GraphQLField<any, any>) {
+    const {resolve = defaultFieldResolver} = field
+    const { settingsType } = this.args
+    field.resolve = async (root, args, ctx, info) => {
+      if (settingsType) {
+        ctx = await getSettings(settingsType, ctx)
```
vwraposo: Do you need to return the ctx? Aren't you changing the ctx object passed?

src/service/worker/runtime/http/middlewares/settings.ts
```diff
@@ -0,0 +1,89 @@
...
+export const getDependenciesHash = (dependencies: AppMetaInfo[]): string => {
+  const dependingApps = pluck('id', dependencies)
+  return createHash('md5')
+    .update(joinIds(dependingApps))
+    .digest('hex')
+}
+
+const formatDependencies = (results: Array<Record<string, any> | undefined>) => {
+  const formatted: any = {}
+  results.forEach(res => {
```
vwraposo: You can use a reduce function

src/service/worker/runtime/graphql/schema/schemaDirectives/Settings.ts
dhasuda: great!

src/service/worker/runtime/graphql/schema/schemaDirectives/Settings.ts
dhasuda: I'll do
```
if (!ctx.vtex) {
    ctx.vtex = {}
  }
  ctx.vtex.settings = { ...ctx.vtex.settings, dependenciesSettings }
```
since ctx is `any` tipe and one request can receive multiple settings. We want to merge them somehow

PR #325, opened 2020-01-31
Faster messages loader

tlgimenes:
Messages loader shared unecessary resposibility with messages client. This lead to a decrease in performance since we had to perform several operations in the messages array both in the dataLoader and in messages client. To address this problem, I've made messages client simple (only making the query) and put all the logic in the dataloader. This lead to an increase in perforamance and less code, since we use many cool features of dataloader, like maxBatching and memoization. 

Since the client has no more built-in logic, if one wants to call messages (like in pages), it should use the 
 messages dataloader instead 

Also, this PR fixes a wild bug in the cache key generator of the http client.

#### Types of changes


src/clients/apps/MessagesGraphQL.ts
```diff
@@ -1,153 +1,152 @@
...
-            indexedByFrom: batch,
-          },
-        },
-      }, {
-        metric: 'messages-translate-v2',
-      })
-      .then(response => response.data!.translateWithDependencies)
-    ))
-    return translations.flat()
+    const response = await this.graphql.query<TranslatedWithDependencies, { args: TranslateWithDependenciesInput }>(
```
vwraposo: Should we remove the batching function here? How could we guarantee that this function will be called only by the dataloader and not any app can call it directly?

tlgimenes: We can't garantee, this is an architectural decision. I prefer to chose having this liability in order to make messages client code simpler

src/clients/apps/MessagesGraphQL.ts
```diff
@@ -1,153 +1,152 @@
...
   srcLang: string
   srcMessage: string
   context?: string
   targetMessage: string
   groupContext?: string
 }
 
-export interface TranslateInputV2 {
-  indexedByFrom: IndexedMessageV2[]
+export interface TranslateInput {
```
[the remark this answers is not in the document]
tlgimenes: :o How does this not erroers ?

PR #334, opened 2020-02-12
Added payment-provider client

marcosvcp:

As title says

Many VTEX Payment apps use this client

#### Types of changes


src/clients/external/PaymentProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+  inbound: (transactionId: string, paymentId: string, action: string) =>
+    `${routes.payment(transactionId, paymentId)}/inbound-request/${action}`,
+}
+
+export class PaymentProvider extends ExternalClient {
+  constructor(protected context: IOContext, options?: InstanceOptions) {
+    super(
+      `http://${context.account}.vtexpayments.com.br/payment-provider`,
+      context,
+      options
```
tlgimenes: why not adding the `x-vtex-use-https` in here instead ? This way we will secure the client for any request

PR #340, opened 2020-02-17
Injects skidder topics into logs to be processed by fluentd

daniel-mitre:

This PR adds **__SKIDDER_TOPIC_1** and **__SKIDDER_TOPIC_2** keys into logs to indicate which topics in *kafka* the logs should be sent by *fluentd*.

We are going to send logs from *fluentd* to *Kafka* (to be consumed by *skidder*) and would be great if the logs arrived in *fluentd* with the topic names, similar to the **SPLUNK_INDEX** injected by *kube-router* in the container metadata section.
I also changed to get the APP_ID from the constants import.

An application should log something with the logger and in *splunk* we should see the log with the new keys.
An app called `vtex.foo` installed in the account `bar` should have the values:
__SKIDDER_TOPIC_1: "skidder.acc.bar"
__SKIDDER_TOPIC_2: "skidder.app.bar.vtex.foo"

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [6.17.0] - 2020-02-18
 ### Added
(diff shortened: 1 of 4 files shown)


daniel-mitre:
I intended to leave this logic for Fluentd, but doing here we would have a lot more flexibility. Also Fluentd administrator would be thankful for logic removed from Fluentd-central. May I implement it here?

PR #363, opened 2020-04-08
Feature/add extensions to graphql client

viniagostini:
I was trying to create a client to make queries to graphql-server, but in order to bypass these queries, we need to add some extensions to it. This PR allows this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [6.25.0] - 2020-04-15
 ### Added
 - Extensions field to graphql client.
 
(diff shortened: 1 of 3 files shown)


src/clients/apps/GraphqlServer.ts
```diff
@@ -0,0 +1,45 @@
...
+import { IOContext } from '../../service/worker/runtime/typings'
+import { IOGraphQLClient } from '../IOGraphQLClient'
+
+const ONE_AND_HALF_SECOND_TIMEOUT = 1500
+
+class GraphqlServerGraphQLClient extends IOGraphQLClient {
+  constructor(context: IOContext, options?: InstanceOptions) {
+    const { account, workspace } = context
+    const name = 'graphql-server'
+    const baseURL = `http://${workspace}--${account}.myvtex.com/_v/private/graphql/v1`
```
brunoabreu, line 15: Why are we using myvtex and not the internal endpoint (app.io.vtex.com)?
Will it always be `/_v/private/`? This path won't let us cache any response.
The `vtex.render-server` already has a `GraphQLServer` client it uses to fetch the `/meta` endpoint. Is it possible to have a single GraphQL Server client?

PR #364, opened 2020-04-17
Change tracing export. Add tracing config to all clients requests

tiagonapoli:
- Add `NODE_ENV` tag to process tracing tags.
- Add tracingConfig to all clients requests.
- Create tracing config on requests to specify if the request span will be child of the root span or will follow from (Refer to `References between Spans` on https://opentracing.io/specification/)
- Create tracing config on requests to specify a suffix to the request span name - it will be `http-request:suffix`
- Change way to export tracing utilities - this way seemed a little off (it is a breaking change, but just vtex.monitoring-proxy is supposed to be using this) 
- Create helper `createTracingContextFromCarrier` to allow worker threads and child_processes to  be traced.

Deploy this version on a test cluster and check for example the spans on `vtex.render-server`

#### Types of changes


gen/manifest.schema
```diff
@@ -41,14 +41,14 @@
                 "metricName": {
                     "type": "string"
```
tiagonapoli, line 42: This file was autogenerated

src/HttpClient/typings.ts
```diff
@@ -4,12 +4,31 @@ import { Span } from 'opentracing'
...
+interface AxiosTracingConfig extends RequestTracingUserConfiguration {
+  tracer: IUserLandTracer
+}
+
+export interface TraceableRequestConfig extends RequestConfig {
+  tracing?: AxiosTracingConfig
+}
+
+export interface RequestTracingConfig {
+  tracing?: RequestTracingUserConfiguration
```
tiagonapoli, line 28: This type, `RequestTracingConfig`, was used on client's methods arguments - It was made in a way it could be replaced with `RequestConfig` in the future, if we want

src/HttpClient/typings.ts
```diff
@@ -4,12 +4,31 @@ import { Span } from 'opentracing'
...
+  rootSpan?: Span
+  referenceType?: SpanReferenceTypes
+  requestSpanNameSuffix?: string
+}
+
+interface AxiosTracingConfig extends RequestTracingUserConfiguration {
+  tracer: IUserLandTracer
+}
+
+export interface TraceableRequestConfig extends RequestConfig {
```
tiagonapoli, line 23: This type was used on `HttpClient` call to axios

src/tracing/UserLandTracer.ts
```diff
@@ -1,11 +1,25 @@
...
   startSpan: Tracer['startSpan']
   inject: Tracer['inject']
   fallbackSpanContext: () => SpanContext
 }
 
+export const createTracingContextFromCarrier = (newSpanName: string, carrier: Record<string, any>): { span: Span, tracer: IUserLandTracer } => {
+  const tracer = TracerSingleton.getTracer()
+  const rootSpan = tracer.extract(FORMAT_HTTP_HEADERS, carrier) as SpanContext | undefined
+  if (rootSpan == null) {
+    throw new Error('Missing span context data on carrier')
```
tiagonapoli, line 14: The user may forget to add the spanContext initiated at the parent process into the carrier

src/tracing/spanReference/SpanReferenceTypes.ts
```diff
@@ -0,0 +1,4 @@
+export enum SpanReferenceTypes {
+    CHILD_OF_REFERENCE = 0,
```
tiagonapoli, line 2: Should we call just `CHILD_OF`? Since the enum name already gives the context on what is this?

src/HttpClient/middlewares/request/setupAxios/interceptors/tracing/index.ts
```diff
@@ -25,8 +25,21 @@ const preRequestInterceptor = (http: AxiosInstance) => (
...
-  const { rootSpan, tracer } = config.tracing
-  const span = tracer.startSpan('http-request', { childOf: rootSpan })
+  const {
+    tracer,
+    rootSpan,
+    requestSpanNameSuffix,
+    referenceType = SpanReferenceTypes.CHILD_OF_REFERENCE,
+  } = config.tracing
+
+  const spanName = requestSpanNameSuffix ? `http-request:${requestSpanNameSuffix}` : 'http-request'
```
Jeymisson, line 35: There are `http-request` spread over your tests. This could be a constant

src/service/worker/runtime/http/middlewares/settings.ts
```diff
@@ -41,12 +42,12 @@ export const getDependenciesHash = (dependencies: AppMetaInfo[]): string => {
     .digest('hex')
 }
 
-export const getDependenciesSettings = async (apps: Apps, assets: Assets) => {
+export const getDependenciesSettings = async (apps: Apps, assets: Assets, tracingConfig?: RequestTracingConfig) => {
   const appId = APP.ID
-  const metaInfos = await apps.getAppsMetaInfos()
+  const metaInfos = await apps.getAppsMetaInfos(undefined, undefined, tracingConfig)
```
Jeymisson, line 47: 😢

src/tracing/spanReference/SpanReferenceTypes.ts
Jeymisson: I'm ok with `CHILD_OF`

src/service/worker/runtime/http/middlewares/settings.ts
tiagonapoli: Sad :(, but using `undefined` will guarantee that the default values will be used

PR #368, opened 2020-05-06
Chore/tracing improvements

tiagonapoli:
- Fix `http.url` entrypoint span tag
- Move ErrorReport parsing logic
- Log request and response headers on entrypoint span
- Fix same ErrorReport, different errorIDs problem
- Log ErrorReport errors to splunk

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- [tracing:ErrorReport] Log ErrorReport errors to splunk.
+
 ## [6.28.3] - 2020-05-15
 ### Fixed
 - [ErrorReport] Update `@vtex/node-error-report` package:
--- a/src/HttpClient/HttpClient.ts
+++ b/src/HttpClient/HttpClient.ts
@@ -36,7 +36,8 @@ type ClientOptions = IOContext & Partial<InstanceOptions>
 export class HttpClient {
   public name: string
 
-  private tracer?: IUserLandTracer
+  private tracer: IOContext['tracer']
+  private logger: IOContext['logger']
   private runMiddlewares: compose.ComposedMiddleware<MiddlewareContext>
 
   public constructor(opts: ClientOptions) {
@@ -70,10 +71,13 @@ export class HttpClient {
       exponentialBackoffCoefficient,
       httpsAgent,
       tracer,
+      logger,
     } = opts
     this.name = name || baseURL || 'unknown'
 
     this.tracer = tracer
+    this.logger = logger
+
     const limit = concurrency && concurrency > 0 && pLimit(concurrency) || undefined
     const headers: Record<string, string> = {
       ...defaultHeaders,
@@ -177,6 +181,7 @@ export class HttpClient {
   protected request = async (config: RequestConfig): Promise<AxiosResponse> => {
     (config as TraceableRequestConfig).tracing = this.tracer ? { 
       ...config.tracing,
+      logger: this.logger,
       tracer: this.tracer,
     } : undefined
 
--- a/src/HttpClient/middlewares/request/setupAxios/__tests__/TracedTestRequest.ts
+++ b/src/HttpClient/middlewares/request/setupAxios/__tests__/TracedTestRequest.ts
@@ -1,6 +1,7 @@
 import { MockReport, MockSpan } from '@tiagonapoli/opentracing-alternate-mock'
 import { AxiosInstance, AxiosResponse } from 'axios'
 import { Span } from 'opentracing'
+import { Logger } from '../../../../../service/logger'
 import { RequestTracingConfig, TraceableRequestConfig } from '../../../../typings'
 import { TestTracer } from './TestTracer'
 
@@ -63,6 +64,13 @@ export class TracedTestRequest {
         ...reqConf,
         tracing: {
           ...reqConf.tracing,
+          logger: new Logger({
+            account: 'mock-account',
+            operationId: 'mock-operation-id',
+            production: false,
+            requestId: 'mock-request-id',
+            workspace: 'mock-workspace',
+          }),
           rootSpan: (this.rootSpan as unknown) as Span,
           tracer: this.tracer,
         },
--- a/src/HttpClient/middlewares/request/setupAxios/interceptors/tracing/index.ts
+++ b/src/HttpClient/middlewares/request/setupAxios/interceptors/tracing/index.ts
@@ -67,7 +67,7 @@ const onResponseError = (err: any) => {
 
   const { requestSpan } = err.config.tracing
   injectResponseInfoOnSpan(requestSpan, err.response)
-  injectErrorOnSpan(requestSpan, err)
+  injectErrorOnSpan(requestSpan, err, err.config.tracing.logger)
   requestSpan.finish()
   return Promise.reject(err)
 }
--- a/src/HttpClient/typings.ts
+++ b/src/HttpClient/typings.ts
@@ -4,6 +4,7 @@ import { Span } from 'opentracing'
 
 import { CacheLayer } from '../caches/CacheLayer'
 import { MetricsAccumulator } from '../metrics/MetricsAccumulator'
+import { IOContext } from '../service/worker/runtime/typings'
 import { SpanReferenceTypes } from '../tracing'
 import { IUserLandTracer } from '../tracing/UserLandTracer'
 import { Cached, CacheType } from './middlewares/cache'
@@ -18,6 +19,7 @@ interface RequestTracingUserConfiguration {
 
 export interface AxiosTracingConfig extends RequestTracingUserConfiguration {
   tracer: IUserLandTracer
+  logger: IOContext['logger']
 }
 
 export interface TraceableRequestConfig extends RequestConfig {
--- a/src/service/tracing/spanSetup.ts
+++ b/src/service/tracing/spanSetup.ts
@@ -1,7 +1,8 @@
 import { Span } from 'opentracing'
 import { ErrorReport } from '../../tracing'
+import { IOContext } from '../worker/runtime/typings'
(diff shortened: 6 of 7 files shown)


src/tracing/errorReporting/utils.test.ts
```diff
@@ -57,7 +61,24 @@ describe('Circular checkings', () => {
       },
       { a: 'abc', b: { d: '12345[...TRUNCATED]' }, c: { d: '12345[...TRUNCATED]' } },
     ],
-  ])(`Doesn't add '[circular]' when references another node but it's not circular %#`, (objCreator, expected) => {
-    expect(truncateStringsFromObject(objCreator(), 5)).toStrictEqual(expected)
+  ] as Array<[() => any, any]>)(
+    `Doesn't add '[circular]' when references another node but it's not circular %#`,
+    (objCreator, expected) => {
+      expect(truncateAndSanitizeStringsFromObject(objCreator(), 5)).toStrictEqual(expected)
```
dhasuda, line 67: I think this line is very long. We can split it:
```
const obj = objCreator()
const result = truncateAndSanitizeStringsFromObject(obj, 5)
expect(result).toStrictEqual(expected)
```

src/tracing/errorReporting/utils.ts
```diff
@@ -1,4 +1,23 @@
-export function truncateStringsFromObject(element: any, maxStrSize: number, depth = 3, seenObjects: any[] = []) {
+const tokenFields = ['authorization', 'authtoken', 'auth', 'token']
+
+export function sanitizeJwtToken(token: string): string {
+  const tokenParts = token.split('.')
+
+  // Not JWT token
+  if (tokenParts.length !== 3) {
```
dhasuda, line 7: Instead of the comment, that can eventually not represent the truth, you can use a function describing it:
```
if (!isJWTToken(tokenParts)) { ... }

const isJWTToken = (tokenParts: string[]) => {
  return tokenParts.length === 3
}
```

src/tracing/errorReporting/utils.ts
```diff
@@ -1,4 +1,23 @@
...
+    return token
+  }
+
+  return `${tokenParts[0]}.${tokenParts[1]}`
+}
+
+export function truncateAndSanitizeStringsFromObject(
```
dhasuda, line 14: Why truncate and sanitize? I think sanitize can embrace the truncate as well. So I suggest `sanitizeStringsFromObject`

src/tracing/errorReporting/utils.ts
```diff
@@ -1,4 +1,23 @@
...
+  element: any,
+  maxStrSize: number,
+  depth = 3,
+  seenObjects: any[] = [],
+  currentKey?: string
+) {
```
dhasuda, line 20: I have a few comments about this, let's go one by one

1. I think it's good to put no more than three parameters in each function signature. I'd split this in two objects:
```
interface SanitizableObject {
  element: any
  metadata: {
    seenObjects: any[] = [],
    currentKey?: string
  }
}

interface SanitizeParams {
  maxStrSize: number
  depth: number
}
```

2. The names `seenObject` and `currentKey` are not intuitive for me. Can you explain some about them for me to get a better sense of it and maybe help in the name?

src/tracing/errorReporting/utils.ts
```diff
@@ -10,8 +29,8 @@ export function truncateStringsFromObject(element: any, maxStrSize: number, dept
       return '[circular]'
```
dhasuda, line 29: I know this is not in the diff but let me be a little annoying 🌝
can we use a constant here?

src/tracing/errorReporting/utils.ts
```diff
@@ -21,15 +40,26 @@ export function truncateStringsFromObject(element: any, maxStrSize: number, dept
     seenObjects.push(element)
     Object.keys(element).forEach(key => {
       // seenObjects.slice creates a copy of seenObjects at the current state
-      element[key] = truncateStringsFromObject(element[key], maxStrSize, depth - 1, seenObjects.slice())
+      element[key] = truncateAndSanitizeStringsFromObject(element[key], maxStrSize, depth - 1, seenObjects.slice(), key)
```
dhasuda, line 43: Now that I see that this `truncateAndSanitizeStringsFromObject` function is being used in a loop, I get a better sense of what is it's purpose. What do you think of creating a class that handles this? Something like this:
```
class SanitizableObject() {
  constructor(private element: any){}

  public sanitized(): any // This would be the loop implementation
  private sanitizeStrings(): any // This would be the `truncateAndSanitizeStringsFromObject`
}
```

src/tracing/errorReporting/utils.ts
```diff
@@ -21,15 +40,26 @@ export function truncateStringsFromObject(element: any, maxStrSize: number, dept
...
     })
 
     return element
-  } else if (objType === 'string') {
+  }
+
+  if (objType === 'string') {
```
dhasuda, line 49: Can we use a switch?
```
switch(objType) {
  case 'object':...
  ...
}
```

src/tracing/errorReporting/errorParsing.ts
```diff
@@ -0,0 +1,53 @@
...
+    requestConfig: {
+      data: requestData,
+      headers: requestHeaders,
+      method,
+      params,
+      timeout: requestTimeout,
+      url,
+    },
+    response: err.response
+      ? {
```
dhasuda, line 37: I'd prefer to make this if before the return. This is kind of personal, but I find it more elegant when the return value is very easy to understand at a glance

src/tracing/errorReporting/ErrorReport.ts
```diff
@@ -56,14 +62,26 @@ export class ErrorReport extends Error {
...
+
   constructor({ kind, message, originalError }: ErrorReportArguments) {
     super(message)
     this.kind = kind
     this.originalError = originalError
-    this.errorId = randomBytes(16).toString('hex')
     this.stack = originalError.stack
     this.message = originalError.message
 
+    if ((originalError as ReportedError).errorReportMetadata) {
```
dhasuda, line 76: I suggest spliting this into a function that clarifies what this `if` means. But, to do that in a more elegant way, you could also transform `ErrorReportArguments ` into a class
```
class ErrorReportArguments {
  constructor(...) {
    ...
    if (alredyReported()) {...} else {...}
    ...
  }

   public alreadyReported() {
     return this.error.errorReportMetadata
  }
}
```

dhasuda: This would eliminate the need to always parse `originalError` to `ReportedError`

src/tracing/errorReporting/ErrorReport.ts
```diff
@@ -108,23 +110,28 @@ export class ErrorReport extends Error {
...
+      if (this.parsedInfo.serverErrorRequestId) {
         span.setTag(TracingTags.ERROR_SERVER_REQUEST_ID, this.parsedInfo.serverErrorRequestId)
       }
     }
 
     span.log({ event: 'error', ...this.toObject() })
+
+    if (!this.alreadyReported && getTraceInfo(span).isSampled) {
+      logger?.error(this.toObject())
+    }
```
dhasuda, line 133: you can separate this check into a function
```
if (mustGenerateLog(this, span)) {...}
const mustGenerateLog (this, span) => {...}
```

src/tracing/errorReporting/ErrorReport.ts
```diff
@@ -108,23 +110,28 @@ export class ErrorReport extends Error {
...
+
     return this
```
dhasuda, line 135: The name of the function `injectOnSpan` is not clear about returning something, I got the impression it would alter values by reference. What about `getSpanWithInjection` or something like this?

tiagonapoli: Unfortunately changing this wouldn't be easy :(
This function is already used by builder-hub and monitoring proxy so we would need to change there first

`this` is returned in this case to allow cascading function calls, for example:

```ts
const tracingSpan: Span
try {
  throw new Error("ASDF")
catch(err) {
  const errReport = ErrorReport.create({
    originalError: err
  }).injectOnSpan(tracingSpan)
}
```

tiagonapoli: This tricky is used in some JS lib, like [chalk](https://www.npmjs.com/package/chalk)

src/tracing/errorReporting/utils.ts
tiagonapoli: This function tries to detect circular js objects - the `seenObjects` holds an array with all objects already seen in the current recursion, so if the next field to visit is already in there this object is circular - for exaple:
```ts
        const obj: any = { a: '123456', b: {} }
        obj.b.c = obj
```

The `currentKey` holds the current object name - in the above example, when analyzing obj.a, the currentKey value would be 'a', this way we can know if the current key is 'auth', or 'authToken' for example

tiagonapoli: This function is based on https://github.com/sindresorhus/serialize-error/blob/master/index.js

tiagonapoli: About changing the arguments, when I was doing this function I thought to do a more ugly/performatic style - this function is recursive with depth limited to 10 currently - and creating an object every new call for the new arguments would cost a little more than just using the arguments

In toolbelt I would fix to use the more elegant way, but in runtime I'm a little afraid to do so

src/tracing/errorReporting/utils.ts
tiagonapoli: I don't like the switch syntax very much :(

In this case I think it would have:
```ts
switch (objType) {
   case 'object':
    ...
   case 'number':
   case 'bigint':
   case 'boolean':
   case 'symbol':
    return element
  case 'function:
    return ...
}
```

tiagonapoli:
The ErrorReport logic was moved to a new [npm package](https://github.com/vtex/node-error-report) , since it was duplicated on node-vtex-api and toolbeot

PR #369, opened 2020-05-11
Use remote sampling

tiagonapoli:
Use jaeger remote sampling. If an error occurs while communicating with the jaeger-agent the sampling fallbacks to probabilistic:0.05 (configurable by the `param` field)

#### Types of changes

### DIFF
--- a/src/service/tracing/TracerSingleton.ts
+++ b/src/service/tracing/TracerSingleton.ts
@@ -33,8 +33,10 @@ export class TracerSingleton {
         agentHost: process.env.VTEX_OWN_NODE_IP,
       },
       sampler: {
+        host: process.env.VTEX_OWN_NODE_IP,
         param: 0.05,
-        type: 'probabilistic',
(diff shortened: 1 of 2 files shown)


src/service/tracing/TracerSingleton.ts
```diff
@@ -33,8 +33,10 @@ export class TracerSingleton {
         agentHost: process.env.VTEX_OWN_NODE_IP,
       },
       sampler: {
+        host: process.env.VTEX_OWN_NODE_IP,
         param: 0.05,
-        type: 'probabilistic',
+        refreshIntervalMs: 60 * 1000,
```
tiagonapoli, line 38: The first refresh is done when the process starts, after `random() * refreshIntervalMs`, to avoid synchronization between processes refreshes.

After the inicial fetch, the refresh is done in intervals (fetching data from the `jaeger-agent` deployed in the host)

PR #374, opened 2020-05-16
Allow more errors per span

tiagonapoli:
Previously error metadata was added to the span tags - this wouldn't allow more than one error ocurring in an span. This was chosen because these tags are indexed for search in jaeger, but log fields are indexed as well, so we can move this data to span logs.

Also added documentation on these log fields.

Send a request with `jaeger-debug-id: anything` header to an endpoint that will error. Search this error on jaeger-ui and check the error log.

#### Types of changes
* [X] Minor improvement

### DIFF
--- a/src/tracing/LogFields.ts
+++ b/src/tracing/LogFields.ts
@@ -3,15 +3,13 @@
 // When using span.log({ field1: string1, field2: object2, ... })
 // The jaeger backend indexes for search the first-level fields that hold strings
 // These fields then can be searched on jaeger-ui using: 'field1=string'
-// This file concentrates all log fieds used in node-vtex-api that hold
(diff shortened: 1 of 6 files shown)


src/tracing/LogFields.ts
```diff
@@ -0,0 +1,24 @@
+/* tslint:disable:object-literal-sort-keys */
+
+// When using span.log({ field1: string1, field2: object2, ... })
+// The jaeger backend indexes for search the first-level fields that hold strings
+// These fields then can be searched on jaeger-ui using: 'field1=string'
+// This file concentrates all log fieds used in node-vtex-api that hold
```
Jeymisson, line 6: all log fields

PR #378, opened 2020-05-19
Add If-Match support to VBase client

athoscouto:
To allow all the creatures of this mighty land to use VBase's `If-Match` feature when developing node apps.

On this PR I basically:
- Introduce `getRawJSON` so people can get a `JSON` as response and the respective headers.
- Add a `ifMatch?: string` param to all save/delete functions of the client.

I know it sucks to add an the extra optional parameter and would be much better to have a "options object", but I didn't want to break compatibility with the current API. If you have any idea of how to have both, please let me know.

People not being able to use `If-Match` logic on node apps.

I created a simple app which receives a `GET` or `PUT` request and proxies it through the VBase client. I've linked `node-vtex-api` with the app and made sure the `If-Match` logic worked.

:not-sure:

#### Types of changes

* [x] Requires change to documentation, which hasn't been updated accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Support to `If-Match` header usage on VBase client.
+
 ### Changed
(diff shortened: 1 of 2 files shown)


Jeymisson:
Our clients have an object called `InstanceOptions` that is passed to the constructor.
This `InstanceOptions` has headers as an option. What do you guys think about putting ifmatch there?

PR #379, opened 2020-05-19
Masterdata client

saviomuniz:

To add Masterdata client to IOClients

Multiple implementations of Masterdata client across the platform.

Can be used through ctx.clients.masterdata

On a handler: 

```
export async function helloworld(ctx: Context, next: () => Promise<any>) {
  ctx.status = 200
  ctx.body = await ctx.clients.masterdata.createOrUpdateEntireDocument('nodevtexapitest', {test: 'test document'})
  await next()
}
```

You also need to add the correct policies to the manifest: 
```
{
  "name": "outbound-access",
  "attrs": {
    "host": "api.vtex.com",
    "path": "/{{account}}/dataentities/*"
  }
},
{
  "name": "ADMIN_DS"
}
```
#### Types of changes

### DIFF
--- a/src/clients/IOClients.ts
+++ b/src/clients/IOClients.ts
@@ -3,6 +3,7 @@ import { IOContext } from '../service/worker/runtime/typings'
 import { Billing, Builder, MessagesGraphQL, Settings } from './apps'
 import { CatalogGraphQL } from './apps/catalogGraphQL/index'
 import { ID, PaymentProvider } from './external'
+import { MasterData } from './external/Masterdata'
 import {
   Apps,
   Assets,
@@ -60,6 +61,10 @@ export class IOClients {
 
   public get licenseManager() {
     return this.getOrSet('licenseManager', LicenseManager)
+  }
+
+  public get masterdata() {
+    return this.getOrSet('masterdata', MasterData)
   }
 
   public get messagesGraphQL() {
--- a/src/clients/IOClients.ts
+++ b/src/clients/IOClients.ts
@@ -2,8 +2,7 @@ import { InstanceOptions } from '../HttpClient'
 import { IOContext } from '../service/worker/runtime/typings'
 import { Billing, Builder, MessagesGraphQL, Settings } from './apps'
 import { CatalogGraphQL } from './apps/catalogGraphQL/index'
-import { ID, PaymentProvider } from './external'
-import { MasterData } from './external/Masterdata'
+import { ID, MasterData, PaymentProvider } from './external'
 import {
   Apps,
   Assets,
--- a/src/clients/external/Masterdata.ts
+++ b/src/clients/external/Masterdata.ts
@@ -1,5 +1,5 @@
 import FormData from 'form-data'
-import { InstanceOptions, IOContext, UserInputError } from '../..'
+import { InstanceOptions, IOContext, UserInputError, RequestTracingConfig } from '../..'
 import { ExternalClient } from './ExternalClient'
 
 export class MasterData extends ExternalClient {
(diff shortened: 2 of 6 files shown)


src/clients/external/Masterdata.ts
```diff
@@ -0,0 +1,170 @@
...
+    pagination: PaginationArgs
+  ) => this.http.get(this.routes.scroll(dataEntity), {
+    headers: paginationArgsToHeaders(pagination),
+    metric: 'masterdata-scrollDocuments',
+    params: {
+      _fields: generateFieldsArg(fields),
+    },
+  })
+
+  public deleteDocument = (dataEntity: string, id: string) =>
```
tiagonapoli, line 132: _performance nit_: Prefer not to use arrow functions in classes (I know that other clients use it, but changing them would be a breaking change, unfortunately):
https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1

src/clients/external/Masterdata.ts
```diff
@@ -0,0 +1,170 @@
...
+    this.http.delete(this.routes.document(dataEntity, id), {
+      metric: 'masterdata-deleteDocument',
```
tiagonapoli, line 134: Can you add tracing options to these methods, just like:
https://github.com/vtex/node-vtex-api/blob/33391b8252a42d41718be7b0a005216c55b23708/src/clients/infra/VBase.ts#L43-L50

tiagonapoli: The `requestSpanNameSuffix` can be the same as the metric

src/clients/external/Masterdata.ts
```diff
@@ -0,0 +1,286 @@
...
+      attachments: (dataEntity: string, id: string, fields: string) =>
+        `${dataEntity}/documents/${id}/${fields}/attachments`,
+      document: (dataEntity: string, id: string) => `${dataEntity}/documents/${id}`,
+      documents: (dataEntity: string) => `${dataEntity}/documents`,
+      publicSchema: (dataEntity: string, schema: string) => `${dataEntity}/schemas/${schema}/public`,
+      schema: (dataEntity: string, schema: string) => `${dataEntity}/schemas/${schema}`,
+      scroll: (dataEntity: string) => `${dataEntity}/scroll`,
+      search: (dataEntity: string) => `${dataEntity}/search`,
+    }
+  }
```
tiagonapoli, line 28: This can be a `private static` property (that's a nit, for me makes sense that routes is a class property other than an object property):
```ts
 private static routes = {
      attachments: (dataEntity: string, id: string, fields: string) =>
        `${dataEntity}/documents/${id}/${fields}/attachments`,
      document: (dataEntity: string, id: string) => `${dataEntity}/documents/${id}`,
      documents: (dataEntity: string) => `${dataEntity}/documents`,
      publicSchema: (dataEntity: string, schema: string) => `${dataEntity}/schemas/${schema}/public`,
      schema: (dataEntity: string, schema: string) => `${dataEntity}/schemas/${schema}`,
      scroll: (dataEntity: string) => `${dataEntity}/scroll`,
      search: (dataEntity: string) => `${dataEntity}/search`,
  }
```
Also, it's not necessary to create the object with the routes every time we call the routes function, we can strip down some operations by doing the above - creating a routes object with each route function

src/clients/external/Masterdata.ts
```diff
@@ -0,0 +1,286 @@
...
+        },
+      }
+    )
+  }
+
+  public updateEntireDocument(
+    dataEntity: string,
+    id: string,
+    fields: object,
+    schema?: string,
```
tiagonapoli, line 148: Just thinking in the future, adding a new argument here for example would be pretty hard (susceptible to creating a breaking change). If we use an object as argument:
```ts
public updateEntrireDocument({ dataEntity, id, fields, schema }: UpdateDocumentArgs, tracingConfig?: RequestTracingConfig) {
```
In the future, if necessary, we could easily add a new optional field that tweaks the behavior.
It's just a note and personal opinion, you don't have to change it - I would prefer to hear more opinions on this

src/clients/external/Masterdata.ts
```diff
@@ -0,0 +1,286 @@
...
+    })
+  }
+
+  public searchDocuments<T>(
+    dataEntity: string,
+    fields: string[],
+    where: string,
+    pagination: PaginationArgs,
+    schema?: string,
+    sort?: string,
```
tiagonapoli, line 190: I think here it would be pretty good to use an object here - I think the developer experience worsens when there are many arguments to fill, or not fill, like:
```ts
masterdata.searchDocument('example', ['123'], 'locale', { page: 1 }, 'wololo', 'increasing')
```
Just peaking the arguments, what does each one represents? 

Now using an object:
```ts
masterdata.searchDocument({
  dataEntity: 'example',
  fields: ['123'],
  where: 'locale',
  pagination: { page: 1 },
  schema: 'wololo',
  sort: 'increasing'
})
```
Now just a peak onto the arguments better contextualizes us with what the call represents

tiagonapoli: And if in the future we want to add another optional argument the user wouldn't be affected at all

src/clients/IOClients.ts
```diff
@@ -62,6 +63,10 @@ export class IOClients {
     return this.getOrSet('licenseManager', LicenseManager)
   }
 
+  public get masterdata() {
+    return this.getOrSet('masterdata', MasterData)
+  }
```
tiagonapoli, line 68: What if the user already implemented a `masterdata` client? Would this override it? That could be a problem :thinking:

src/clients/external/Masterdata.ts
```diff
@@ -0,0 +1,286 @@
+import FormData from 'form-data'
+import { InstanceOptions, IOContext, RequestTracingConfig, UserInputError } from '../..'
+import { ExternalClient } from './ExternalClient'
+
+export class MasterData extends ExternalClient {
+  public constructor(ctx: IOContext, options?: InstanceOptions) {
+    super(`http://api.vtex.com/${ctx.account}/dataentities`, ctx, {
+      ...options,
+      headers: {
+        ...(options && options.headers),
```
tiagonapoli, line 10: _nit_: This can be:
```ts
...options?.headers
```

src/clients/external/Masterdata.ts
```diff
@@ -0,0 +1,286 @@
...
+        ...{ Accept: 'application/vnd.vtex.ds.v10+json' },
```
tiagonapoli, line 11: Why using destructuring here instead of:
```ts
Accept: 'application/vnd.vtex.ds.v10+json'
```

src/clients/IOClients.ts
saviomuniz: Nope, the internal clients extends the IOClients, therefore overriding it.

src/clients/external/Masterdata.ts
```diff
@@ -0,0 +1,286 @@
+import FormData from 'form-data'
+import { InstanceOptions, IOContext, RequestTracingConfig, UserInputError } from '../..'
+import { ExternalClient } from './ExternalClient'
+
+export class MasterData extends ExternalClient {
+  public constructor(ctx: IOContext, options?: InstanceOptions) {
+    super(`http://api.vtex.com/${ctx.account}/dataentities`, ctx, {
```
ygorops, line 7: @saviomuniz passa o account no `an` ao invés do primeiro path.

src/clients/external/Masterdata.ts
```diff
@@ -0,0 +1,286 @@
...
+        ...(options && options.headers),
+        ...{ Accept: 'application/vnd.vtex.ds.v10+json' },
+        VtexIdclientAutCookie: ctx.authToken,
+      },
+    })
+  }
+
+  private get routes() {
+    return {
+      attachments: (dataEntity: string, id: string, fields: string) =>
```
ygorops, line 19: A API de attachments não funciona para o MD V2.

src/clients/external/Masterdata.ts
```diff
@@ -0,0 +1,300 @@
...
+  public constructor(ctx: IOContext, options?: InstanceOptions) {
+    super(`http://api.vtex.com/api/dataentities`, ctx, {
+      ...options,
+      headers: {
+        Accept: 'application/vnd.vtex.ds.v10+json',
+        VtexIdclientAutCookie: ctx.authToken,
+        ...options?.headers,
+      },
+      params: {
+        an: ctx.account,
```
tiagonapoli, line 23: Should we allow the user to set another params options? I don't know if it makes sense

src/clients/IOClients.ts
```diff
@@ -3,6 +3,7 @@ import { IOContext } from '../service/worker/runtime/typings'
 import { Billing, Builder, MessagesGraphQL, Settings } from './apps'
 import { CatalogGraphQL } from './apps/catalogGraphQL/index'
 import { ID, PaymentProvider } from './external'
+import { MasterData } from './external/Masterdata'
```
tiagonapoli, line 6:
```suggestion
import { ID, MasterData, PaymentProvider } from './external'
```

src/clients/external/Masterdata.ts
```diff
@@ -0,0 +1,302 @@
...
+  scroll: (dataEntity: string) => `${dataEntity}/scroll`,
+  search: (dataEntity: string) => `${dataEntity}/search`,
+}
+
+export class MasterData extends ExternalClient {
+  public constructor(ctx: IOContext, options?: InstanceOptions) {
+    super(`http://api.vtex.com/api/dataentities`, ctx, {
+      ...options,
+      headers: {
+        Accept: 'application/vnd.vtex.ds.v10+json',
```
ygorops, line 18: Here you can use `application/json`.

PR #383, opened 2020-05-21
Fix/masterdata client issues

saviomuniz:

To fix masterdata client sort type and scroll api

Sort is not an enum, it's rather a string field that you have to say which field you want to sort by and the order (e.g. `count DESC`) 🤦🏽‍♂️, also there was a problem on scroll api documentation, and pagination was wrong

Use `searchDocuments` or `scrolDocuments` endpoint of the client

```
 masterdata.scrollDocuments(
    {
      dataEntity: COURSE_ENTITY,
      fields: ['count', 'slug'],
      size: 4,
      token: '{your_token}',
      schema: 'v1',
      sort: 'count DESC',
    }
  )
```

#### Types of changes

### DIFF
--- a/src/clients/external/Masterdata.ts
+++ b/src/clients/external/Masterdata.ts
@@ -183,7 +183,7 @@ export class MasterData extends ExternalClient {
   }
 
   public scrollDocuments<T>(
-    { dataEntity, fields, schema, size, sort, token }: ScrollInput,
+    { dataEntity, fields, mdToken, schema, size, sort, }: ScrollInput,
     tracingConfig?: RequestTracingConfig
   ) {
     const metric = 'masterdata-scrollDocuments'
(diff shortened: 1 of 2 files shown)


src/clients/external/Masterdata.ts
```diff
@@ -183,23 +183,24 @@ export class MasterData extends ExternalClient {
...
+        _size: size,
         _sort: sort,
+        _token: token,
       },
       tracing: {
         requestSpanNameSuffix: metric,
         ...tracingConfig?.tracing,
       },
-    })
+    }).then(({headers: {'x-vtex-md-token': resToken}, data}) => ({token: resToken, data}))
```
tiagonapoli, line 203: Should we return `{ masterdataToken, data }` instead? So users doesn't mix the semantics of this token with the app token on `ctx.vtex`?

PR #388, opened 2020-06-04
Add missing fields in the type AppMetaInfo

Klynger:
Add missing fields of `AppMetaInfo`

The type `AppMetaInfo` doesn't have all the fields that it is returning

Here is a print of a call to `ctx.clients.apps.getAppsMetaInfos()`
image

#### Types of changes

### DIFF
--- a/src/clients/infra/Apps.ts
+++ b/src/clients/infra/Apps.ts
@@ -116,11 +116,11 @@ export class Apps extends InfraClient {
       return this.installRuntime(descriptor, tracingConfig)
     }
 
-    const metric = 'apps-install' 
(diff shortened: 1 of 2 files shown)


src/clients/infra/Apps.ts
```diff
@@ -595,7 +595,13 @@ interface ZipOptions {
 
 export interface AppMetaInfo {
   id: string
+  name: string
+  credentialType: string
+  registry: string
+  vendor: string
+  version: string
   settingsSchema?: Record<string, any>
+  _activationDate: string
```
Klynger, line 604: These are the only changes that I made, the rest is just lint

PR #391, opened 2020-06-12
Fix ErrorReport create

tiagonapoli:

The the ErrorReport.create is now implemented it's possible to create recursive ErrorReport, e.g.:

```ts
const a = new Error('asdf')
const b = ErrorReport.create({ originalError: a })
const c = ErrorReport.create({ originalError: b }) 
```
The originalError is supposed to be something that's _not_ already an errorReport, so when using a `catch-all` code and wrapping the error with an errorReport, we had to check if the error to be wrapped wasn't already an ErrorReport. Because of this a check on `ErrorReport.create` was added - in case the provided `originalError` is already a `ErrorReport` it just returns. 

#### Types of changes


src/tracing/errorReporting/ErrorReport.ts
```diff
@@ -12,10 +12,37 @@ import { LOG_FIELDS } from '../LogFields'
...
+   * In case the args.originalError argument is already an ErrorReport
+   * instance, then ErrorReport.create just returns it. If it's not,
+   * it returns a new ErrorReport wrapping the error. This way you can
+   * use ErrorReport.create on a catchAll, e.g.:
+   *
+   * ```
+   * try {
+   *   await next()
+   * } catch(err) {
+   *   ErrorReport.create({ originalError: err }).injectOnSpan(span)
```
Jeymisson, line 27: It looks much better this way. One consideration, shouldn't this relation be on the span. I mean, shouldn't we call `span.injectErrorReport(errorReport)` rather than the opposite? I believe that's what is done inside the `injectOnSpan` method.

Jeymisson: Imagine that every lib we have could have an inject method like this. This way all libs should know our span structure, if we change our span structure, don't want jaeger anymore, we should change all libs. If this abstraction goes somewhere else, a span wrapper, we can change one place only.

tiagonapoli: The current Spans, SpanContexts and Tracer follow the OpenTracingAPI, which was created to solve the problem you mentioned (what if we don't want jaeger anymore? We would have to reinstrument everything?) -- If we change our distributed tracing backend to another OpenTracing compatible backend we would have to change just the exporter, which is this piece of code [here](https://github.com/vtex/node-vtex-api/blob/e58722ac062a60ed482893293fb1df488cd13d9d/src/service/tracing/TracerSingleton.ts#L30). However if we change from OpenTracing to OpenTelemetry, which we will certainly want, because Opentelemetry is the next major of OpenTracing and will ship with many things that will help us a lot, then we would have to do more changes :(

Good for us, there's already an official implementation of a [shim](https://github.com/open-telemetry/opentelemetry-js/tree/master/packages/opentelemetry-shim-opentracing), which wraps the OpenTelemetry API and provides the OpenTracing API, so changing to OpenTelemetry wouldn't be a problem.

If we had a Span wrapper ideally we would have to implement something alike to this shim, so it would be doable, and I think it's a good idea -- we would be prepared already for OpenTelemetry.

tiagonapoli: For now however I think it's ok to use `ErrorReport.create`, until we implement this shim. In the near future I will design a plan to implement it

PR #392, opened 2020-06-13
Remove unnecessary span

tiagonapoli:

In the initial runtime native instrumentation a span wrapping the user middlewares was created, however no useful tags or logs was added to it, so I think it just added complexity to the traces, by adding an additional parent-child relationship. The useful thing about this span was to observe the runtime's middlewares overhead, but this could be solved by adding logs to the entrypoint span - and this overhead in many times was negligble compared to the full tracing, so the entrypoint span and the span wrapping the user middlewares often appeared to have almost the same sizes:

Screenshot from 2020-06-12 20-54-18

Screenshot from 2020-06-12 20-54-28

Screenshot from 2020-06-12 20-54-47

This PR proposes to remove this span and add logs onto the entrypoint span marking the moments in which the user middlewares started and finished:

Screenshot from 2020-06-12 20-59-26

The runtime overhead is easily calculated by subtracting the `duration-ms` from the entrypoint span duration.

#### Types of changes
* [X] Refactor

### DIFF
--- a/src/service/tracing/tracingMiddlewares.ts
+++ b/src/service/tracing/tracingMiddlewares.ts
@@ -1,8 +1,10 @@
 import { FORMAT_HTTP_HEADERS, SpanContext, Tracer } from 'opentracing'
 import { ACCOUNT_HEADER, REQUEST_ID_HEADER, TRACE_ID_HEADER, WORKSPACE_HEADER } from '../../constants'
 import { ErrorReport, getTraceInfo } from '../../tracing'
+import { LOG_EVENTS } from '../../tracing/LogFields'
 import { Tags } from '../../tracing/Tags'
 import { UserLandTracer } from '../../tracing/UserLandTracer'
+import { hrToMillis } from '../../utils'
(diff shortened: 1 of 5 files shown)


src/service/tracing/tracingMiddlewares.ts
```diff
@@ -57,24 +59,26 @@ export const nameSpanOperationMiddleware = (operationType: string, operationName
...
+    const startTime = process.hrtime()
 
     try {
+      span.log({ event: LOG_EVENTS.USER_MIDDLEWARES_START })
       await next()
     } catch (err) {
       ErrorReport.create({ originalError: err }).injectOnSpan(span, ctx.vtex.logger)
       throw err
     } finally {
+      span.log({ event: LOG_EVENTS.USER_MIDDLEWARES_FINISH, 'duration-ms': hrToMillis(process.hrtime(startTime)) })
```
Jeymisson, line 80: I'd just extract `duration-ms` to a constant.

PR #394, opened 2020-06-22
Improve HTTP request tracing

tiagonapoli:

This PR creates a span started at the client level (before running axios) that encloses all requests done by a call to a client. 

Right now I think retries on the same request are kinda difficult to spot because they are all mixed up with other requests. The new span this PR creates makes it easier to spot these cases, and also logs info on the retries and the wait time until the request will be retried on a `event: retryable-request-fail`.

The span created enclosing the retries on a request:
Screenshot from 2020-06-22 11-43-02

Here we can see the `event: retryable-request-fail` log:
Screenshot from 2020-06-22 11-43-54

Unfortunately, when the request is successful right away, this span adds one more layer of complexity. To get more use on this span this PR added information on the caching middlewares implemented in the clients.

We now have tags specifying cache strategies enabled for the request and if we have miss or hit in a cache type:
Screenshot from 2020-06-22 11-38-03

And we have logs on some cache middlewares events:
Screenshot from 2020-06-22 11-38-13

You can test all this linking the app on this repo, and following the instructions on readme (make sure to do the requests with jaeger-debug-id header):
https://github.com/tiagonapoli/vtex-io-cache-examples

You can deploy `service-runtime-node@6.24.0-beta.5` on a test cluster to test this PR as well, it will sample 100%. 
https://vtex.slack.com/archives/CMJU6HLGG/p1592935074045500

#### Types of changes

### DIFF
--- a/src/HttpClient/middlewares/cache.ts
+++ b/src/HttpClient/middlewares/cache.ts
@@ -91,20 +91,17 @@ export const cacheMiddleware = ({ type, storage }: CacheOptions) => {
     }
 
     const span = ctx.tracing!.rootSpan
-    const isTraceSampled = ctx.tracing!.isSampled
 
     const key = cacheKey(ctx.config)
     const segmentToken = ctx.config.headers[SEGMENT_HEADER]
     const keyWithSegment = key + segmentToken
 
-    if(isTraceSampled) {
-      span.log({
-        event: HttpLogEvents.CACHE_KEY_CREATE,
-        [HttpCacheLogFields.CACHE_TYPE]: cacheType,
-        [HttpCacheLogFields.KEY]: key,
-        [HttpCacheLogFields.KEY_WITH_SEGMENT]: keyWithSegment,
-      })
-    }
+    span.log({
+      event: HttpLogEvents.CACHE_KEY_CREATE,
+      [HttpCacheLogFields.CACHE_TYPE]: cacheType,
+      [HttpCacheLogFields.KEY]: key,
+      [HttpCacheLogFields.KEY_WITH_SEGMENT]: keyWithSegment,
+    })
 
     const cacheHasWithSegment = await storage.has(keyWithSegment)
     const cached = cacheHasWithSegment ? await storage.get(keyWithSegment) : await storage.get(key)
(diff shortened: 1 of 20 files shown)


src/HttpClient/middlewares/cache.ts
```diff
@@ -60,44 +63,92 @@ export enum CacheType {
...
       const {etag: cachedEtag, response, expiration, responseType, responseEncoding} = cached as Cached
+
       if (type === CacheType.Disk && responseType === 'arraybuffer') {
         response.data = Buffer.from(response.data, responseEncoding)
       }
 
-      if (expiration > Date.now()) {
+      const now = Date.now()
+
+      if (isTraceSampled) {
```
Jeymisson, line 121: As this if is repeated, I think we would like to have something like `span.logIfSampled(...)` in our span abstraction. What do you think?

src/HttpClient/middlewares/tracing.ts
```diff
@@ -0,0 +1,80 @@
...
+      logger,
+      rootSpan: span,
+      tracer,
+    }
+
+    // We can't pass ctx.tracing to axios, so we pass ctx.config.tracing with
+    // a reference to the ctx.tracing object (axios only receives the config object)
+    ;(ctx.config as TraceableRequestConfig).tracing = ctx.tracing
+
+    const hasMemoCache = !(!isLocallyCacheable(ctx.config, CacheType.Any) || !ctx.config.memoizable)
```
Jeymisson, line 48: !(!isLocallyCacheable(ctx.config, CacheType.Any) || !ctx.config.memoizable)
`ctx.config.memoizable = false` => `!(... || !ctx.config.memoizable )` = true is this ok?

src/service/tracing/TracerSingleton.ts
```diff
@@ -34,9 +34,9 @@ export class TracerSingleton {
       },
       sampler: {
         host: process.env.VTEX_OWN_NODE_IP,
-        param: 0.05,
+        param: 1,
```
Jeymisson, line 37: 🔨

src/HttpClient/middlewares/cache.ts
tiagonapoli: Actually, `span.log` already does this check. There's an `if(this.writable)` in the beginning of almost all span functions so the function does nothing if the trace is not sampled or the span is finished.
https://github.com/jaegertracing/jaeger-client-node/blob/4dd7b3e8d2282241e8be55203331c13eb8208028/src/span.js#L274

But I thought that in relatively low level code as the clients are, we could optimize it a little more to avoid the function call and log object allocation - I was wrong. I did a simple benchmark and maybe we wouldn't gain nothing with this - so this was a sad case of a premature optimization, sorry about that.

I will remove these if's, since they aren't necessary

src/service/tracing/TracerSingleton.ts
tiagonapoli: Will remove this before releasing :D

PR #397, opened 2020-07-02
New Prometheus Metrics

tiagonapoli:

Create new prometheus exported metrics:
- runtime_http_requests_total
The total number of incoming requests, labeled by response status and route handler name (e.g: private-handler:render on vtex.render-server).

- runtime_http_aborted_requests_total: 
The total number requests aborted by the caller (for example, the router aborting a request because of a gateway timeout)

- runtime_http_requests_duration_milliseconds
A Histogram with predefined buckets for requests durations. Prometheus histogram buckets are interesting because they make possible to aggregate multiple pods bucket counts in a way that maintains statistic correctness, however buckets are a little costly because each new bucket creates a new time series (same way with labels, each new label configuration creates a new timeseries), so at first I created 10 buckets with exponentially increasing values. You can check more on histograms [here](https://grafana.com/blog/2020/06/23/how-to-visualize-prometheus-histograms-in-grafana/)

- runtime_http_response_size_bytes
A Histogram for the response sizes. This only counts response sizes for requests that set the `content-lenght` response header, so streamed responses won't be counted.

With these metrics no sampling will occur, we'll get all the population data. (in the case of histograms we will get all the data, but grouped by bucket values)

#### Types of changes


package.json
```diff
@@ -76,8 +76,7 @@
     "mime-types": "^2.1.12",
     "opentracing": "^0.14.4",
     "p-limit": "^2.2.0",
-    "prom-client": "^11.5.3",
-    "prometheus-gc-stats": "^0.6.2",
+    "prom-client": "^12.0.0",
```
tiagonapoli, line 79: prom-client@12.x included the gc-stats

src/service/worker/runtime/builtIn/middlewares.ts
```diff
@@ -1,5 +1,4 @@
-import { collectDefaultMetrics, Gauge, Registry } from 'prom-client'
-import gcStats from 'prometheus-gc-stats'
```
Jeymisson, line 2: You've removed gc-stats, this will remove our garbage collector metrics. Why?

tiagonapoli: I updated the `prom-client` package. The new major has native gc-stats

tiagonapoli: https://github.com/siimon/prom-client/blob/master/CHANGELOG.md#added-1

Jeymisson: Nice!

PR #405, opened 2020-07-24
Adding timestamp field to billing log

lucis:
As of now, our VTEX IO Billing log **doesn't** add a timestamp, what makes it really hard to use the data on Elastic Search.

Elasticsearch used to support automatically adding timestamps to documents being indexed, but [deprecated this feature](https://www.elastic.co/guide/en/elasticsearch/reference/2.0/mapping-timestamp-field.html) in 2.0.0

### DIFF
--- a/src/service/worker/runtime/http/middlewares/timings.ts
+++ b/src/service/worker/runtime/http/middlewares/timings.ts
@@ -42,13 +42,13 @@ const logBillingInfo = (
   millis: number
 ) => JSON.stringify({
   '__VTEX_IO_BILLING': 'true',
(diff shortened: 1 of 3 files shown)


src/service/worker/runtime/http/middlewares/timings.ts
```diff
@@ -48,6 +48,7 @@ const logBillingInfo = (
   'isLink': LINKED,
   'production': production,
   'routeType': type === 'public' ? 'public_route' : 'private_route',
+  'timestamp': new Date().getTime(),
```
tiagonapoli, line 51: This can be `Date.now()`

PR #406, opened 2020-07-30
Add between scrapes event loop metrics

tiagonapoli:
Add new event loop lag metrics - max, p99 and p95 - measured between prometheus scraps:
  - runtime_event_loop_lag_max_between_scrapes_seconds
  - runtime_event_loop_lag_percentiles_between_scrapes_seconds 

#### Types of changes

### DIFF
--- a/src/service/tracing/metrics/measurers/EventLoopLagMeasurer.ts
+++ b/src/service/tracing/metrics/measurers/EventLoopLagMeasurer.ts
@@ -1,11 +1,8 @@
 import { EventLoopDelayMonitor, monitorEventLoopDelay } from 'perf_hooks'
 import { Gauge } from 'prom-client'
-import { promisify } from 'util'
 import { nanosecondsToSeconds } from '../../../../utils'
 import { createEventLoopLagMaxInstrument, createEventLoopLagPercentilesInstrument } from '../instruments'
 import { EventLoopMetricLabels } from '../MetricNames'
-
-const sleep = promisify(setTimeout)
 
 export class EventLoopLagMeasurer {
   private eventLoopDelayMonitor: EventLoopDelayMonitor
@@ -24,7 +21,6 @@ export class EventLoopLagMeasurer {
   }
 
   public async updateInstrumentsAndReset() {
-    await sleep(0)
     this.maxInstrument.set(nanosecondsToSeconds(this.eventLoopDelayMonitor.max))
     this.setPercentileObservation(95)
     this.setPercentileObservation(99)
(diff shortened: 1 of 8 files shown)


src/service/tracing/metrics/measurers/EventLoopLagMeasurer.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.percentilesInstrument = createEventLoopLagPercentilesInstrument()
+    this.maxInstrument = createEventLoopLagMaxInstrument()
+  }
+
+  public start() {
+    this.eventLoopDelayMonitor.enable()
+  }
+
+  public async updateInstrumentsAndReset() {
+    await sleep(0)
```
Jeymisson, line 27: 🤔

Jeymisson: Just to see the time of consuming a promise that does nothing?

Jeymisson: Is sleep the best way of doing this?

tiagonapoli: This forces the event loop to do one last loop before we get the data on the event loop histogram. But thinking now I don't know if that would be necessary

tiagonapoli: In this sample is a promisified version of setTimeout - it adds a callback on the libuv's timeout queue, so the callback will be resolved only when the event loop turns again - since the timeout is zero the next event loop turn will already resolve this sleep promise

tiagonapoli: That would be necessary in case we had this situation:
```js
// start measuring event loop lag
const h = startMonitoring()
h.enable()
// do lots of only synchronous work, without letting the event loop to loop

h.disable()
// this would print 0, even though we blocked the event loop a lot
console.log(h.max)
```

Now, if we did like this:
```js
// start measuring event loop lag
const h = startMonitoring()
h.enable()
// do lots of only synchronous work, without letting the event loop to loop

await promisify(setTimeout)(0)
h.disable()
// this would print the event loop lag on the synchronous work we just did
// now it works because we let the event loop run at least once
console.log(h.max)
```

tiagonapoli: I though of letting the event loop run once just to be safe

PR #411, opened 2020-08-18
Wait for in-flight requests before killing workers

marcosvcp:
Change the worker's SIGTERM handler to wait for service timeout before killing any process.

We have a hypothesis that workers are killed prematurely without finishing in-flight requests on the application lifecycle. Our thesis is based on two main points:

1) [Errors happening when pods are scaling down/removed](https://vtex.slack.com/archives/CCYP7AS59/p1597421846043800?thread_ts=1597261439.002200&cid=CCYP7AS59)
2) [The way that Kubernetes handle a graceful shutdown](https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html)

We notice that currently, our timeout to [kill any process running on the pod is 1.5s](https://github.com/vtex/node-vtex-api/blob/master/src/service/master.ts#L74) after receives SIGTERM.

Based on that we are changing the following rules:

1) Since our services timeout can be set up to 60s, the process will wait at max the specified timeout.
2) The worker will wait for at least 30s before sending SIGKILL to child workers and then, exit their own process, this time is based on Kubernetes default time.

We expect to won't see this problem happening again since all in-flight requests can be finished up before the SIGKILL are sent.

Also, we set SIGINT to 1s, why?

image
[Source](https://man7.org/linux/man-pages/man7/signal.7.html)

We don't have any production usage of SIGINT, so we set to 1s just in case you need to run - in dev time - the runtime on your own machine.

_There's another change pending on platform to set the time to send SIGKILL based on service time._

#### Types of changes

### DIFF
--- a/src/service/master.ts
+++ b/src/service/master.ts
@@ -4,11 +4,7 @@ import { constants } from 'os'
 import { INSPECT_DEBUGGER_PORT, LINKED, UP_SIGNAL } from '../constants'
 import { isLog, logOnceToDevConsole } from './logger'
 import { logger } from './worker/listeners'
-import {
-  broadcastStatusTrack,
-  isStatusTrackBroadcast,
-  trackStatus,
-} from './worker/runtime/statusTrack'
+import { broadcastStatusTrack, isStatusTrackBroadcast, trackStatus } from './worker/runtime/statusTrack'
 import { ServiceJSON } from './worker/runtime/typings'
 
 let handledSignal: NodeJS.Signals | undefined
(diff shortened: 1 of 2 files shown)


src/service/master.ts
```diff
@@ -56,45 +50,53 @@ const onOnline = (worker: Worker) => {
   }
 }
 
-const handleSignal: NodeJS.SignalsListener = signal => {
+const GRACEFULLY_SHUTDOWN_TIMEOUT_S = 30
+const SIGINT_TIMEOUT_S = 1
```
tlgimenes, line 54: why sigint is only one second in here ? Won't we end up falling in the same problem ?

marcosvcp: Sorry, I changed the PR description to cover that.

SIGINT is a keyboard interruption, also [it is a sort of "force" to kill signal](https://stackoverflow.com/questions/4042201/how-does-sigint-relate-to-the-other-termination-signals-such-as-sigterm-sigquit). (Like CTRL+C)

PR #419, opened 2020-09-10
Fix images on docs

tiagonapoli:
git was changing the .png binaries, corrupting them. This was solved by changing the .gitattributes file to add an exception to *.png files


VerasThiago (approved):
LGTM

PR #426, opened 2020-11-13
Fix node cluster workers heap size

victorges:
We started using `cluster` in the node runtime but forgot to
tune the forked processes heap sizes so that they wouldn't
each of them try to use all the memory that was available
in the container. This is measurable in our most memory-intensive
apps which are now currently operating way above their requested
memory.

This change fixes it by tuning the `execArgv` of the worker processes
to split the memory available between them, so the `node` runtime
tries to keep them below that threshold.

Memory abuse by most memory-intensive apps.
e.g. [vtex.render-to-string](http://grafana.ingress.vtex.io/d/0FbY1kRWk/io-deployments?viewPanel=18&orgId=1&var-datasource=prometheus%20stores-2a&var-deployment_namespace=vendor-vtex&var-deployment_name=vtex-render-to-string-1-4-0-42b9058a3b29399)

Deploy runtime with this change in a cluster and make sure that apps
stay below their requested memory.

Note: I'm thinking about tuning the `max-old-space-size` that we set in the runtime
so it's larger than 75% of requested memory, as we can probably use more than
that since we run in containers not VMs.

#### Types of changes


src/service/master.ts
```diff
@@ -75,17 +75,39 @@ const handleSignal = (timeout: number): NodeJS.SignalsListener => (signal) => {
...
   const { workers: numWorkers, timeout = GRACEFULLY_SHUTDOWN_TIMEOUT_S } = service
 
   if (service.deterministicVary) {
     process.env.DETERMINISTIC_VARY = 'true'
   }
 
-  // Setup dubugger
+  const settings: ClusterSettings = {
+    execArgv: transformArg(process.execArgv, 'max-old-space-size', size => Math.round(parseInt(size, 10)/numWorkers)),
+  }
```
athoscouto, line 105: What about this?

```suggestion
  const settings: ClusterSettings = {
    execArgv: transformArg(process.execArgv, 'max-old-space-size', size => Math.round(parseInt(size, 10)/numWorkers)),
    // Setup debugger
    ...(LINKED ? { inspectPort: INSPECT_DEBUGGER_PORT } : {}),
  }
  
```

If you don't like it, at least consider the new line after the code block 😄

athoscouto (approved):
LGTM

PR #443, opened 2021-03-16
Updated workspaceFields

bentoper:
Required for [this](https://github.com/vtex/colossus/pull/320) PR.


athoscouto:
This ship has sailed!

More than a year ago we decided to break compatibility to avoid fetching too much data from apps.

PR #446, opened 2021-04-20
ICS-38 Feat/rate limit

VerasThiago:

Enable user to config the amount of requests to be processed per minute and concurrently by setting it in `service.json`.

Example:

```json
{
  "memory": 256,
  "ttl": 10,
  "timeout": 2,
  "minReplicas": 2,
  "maxReplicas": 4,
  "workers": 1,
  "globalRateLimit": {
    "perMinute": 4,
    "concurrent": 20
  },
  "events": {
    "depsUpdate": {
      "keys": ["test-event-veras"],
      "rateLimitPerReplica": {
        "perMinute": 4,
        "concurrent": 2
      }
    },
    "workspacePromotion": {
      "keys": ["test-event-veras2"], 
      "rateLimitPerReplica": {
	"perMinute": 8,
        "concurrent": 10
       }
    }
  }, 
  "routes": {
    "installations": {
      "path": "/installations/:appId",
      "rateLimitPerReplica": {
        "perMinute": 3,
        "concurrent": 4
      }
    },
    "account": {
      "path": "/installations/:appId/:accountName"
    }
  }
}
```

We are having constant crises with apps not being able to process an increased amount of events. We believe that the simplest solution for that is to provide a simple way for apps to rate-limit the amount of events that they process concurrently/in a time period. 

That way, they would process only a configured amount of events that the developer know the app can handle and return 429 errors to the additional events, so Courier can retry the delivery later after the exponential back-off. 

Link this lib into `/node` folder of some `vtex` app that have `events` or `routes`.

Set the rate limit inside `service.json` using the example above.

Send requests to `courier` or the app `route` to force `429` error.

image

#### Types of changes

* [x] Requires change to documentation, which has been updated accordingly.


src/service/worker/runtime/http/index.ts
```diff
@@ -1,5 +1,6 @@
 import { IOClients } from '../../../../clients/IOClients'
 import { nameSpanOperationMiddleware, traceUserLandRemainingPipelineMiddleware } from '../../../tracing/tracingMiddlewares'
+import { concurrentRateLimiter, perMinuteRateLimiter } from '../events/middlewares/rateLimit'
```
victorges, line 3: The middlewares should proably not be in the `/events/` folder, since they are used across the whole pipeline now.

src/service/worker/runtime/typings.ts
```diff
@@ -184,6 +191,7 @@ export interface RawServiceJSON {
   timeout?: number,
   runtimeArgs?: string[],
   routes?: Record<string, ServiceRoute>,
+  globalRateLimit?: Record<string, RateLimiter>
```
victorges, line 194: I think we should have well-defined supported keys here, only `events` and `routes`.

Maybe this would work? 
```suggestion
  globalRateLimit?: Pick<Record<string, RateLimiter>, 'events' | 'routes'>
```

If not, we may need to declare a custom `GlobalRateLimiter` interface which has only the actual fields we support.

victorges: Also just not to forget, reminder that the global rate limiter still needs to be plugged in the pipeline

src/typings/tokenbucket.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'tokenbucket'
```
victorges, line 1: Is this the necessary compatibility for the coffescript types?

victorges (commented):
LGTM!

Lots of comments since I generally like to iterate on the lower levels of the code for future maintainability, and cause you said you welcomed some discussions 😄 

Not giving the approval yet only cause I want to see the implementation of the global rate limiter when it's done, just ping me so I can take a look here.

src/service/worker/runtime/events/middlewares/rateLimit.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export function perMinuteRateLimiter(rateLimit: RateLimiter | undefined) {
+  if (!rateLimit || !rateLimit.perMinute) {
+    return noopMiddlewareFunc()
+  }
+
+  const tokenBucket = new TokenBucket({
+    interval: 'minute',
+    size: rateLimit.perMinute,
```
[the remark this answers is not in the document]
victorges: Oops, also remembered of a **required** change: We must divide these limits by the number of services workers as well. Possibly `ceil`ing the result to avoid any bad truncations like to `0`.

To avoid receiving big unrelated configuration objects here on these middleware creation functions, consider receiving only the number of workers as a second argument, or maybe even only the desired rate limit as a single `number` argument. So the callers of these functions can do any adjustment themselves based on the number of `RawServiceJSON.workers`. I like this last idea, since it keeps these middlewares very single purpose, not handling any context-related logic regarding parsing and interpreting the service JSON objects.

src/service/worker/runtime/typings.ts
```diff
@@ -184,6 +191,7 @@ export interface RawServiceJSON {
   timeout?: number,
   runtimeArgs?: string[],
   routes?: Record<string, ServiceRoute>,
+  globalRateLimit?: RateLimiter
```
victorges, line 194: We should rename this to just `rateLimitPerReplica` as well, so it is not confused with a really "global" rate limit that would limit across all replicas of the service.

src/service/worker/runtime/typings.ts
```diff
@@ -167,14 +167,21 @@ export interface ServiceRoute {
...
 
 export interface ServiceEvent {
   keys?: string[],
   sender?: string,
   subject?: string,
   settingsType?: RouteSettingsType,
+  rateLimitPerReplica?: RateLimiter
+}
+
+export interface RateLimiter {
```
victorges, line 182: **nit:** WDYT of another string that matches this regex instead? `RateLimit(er)?(Config|Options)?`

src/service/worker/index.ts
```diff
@@ -87,7 +91,8 @@ interface HttpHandlerByScope {
 
 const createAppHttpHandlers = (
   { config: { routes, clients } }: Service<IOClients, RecorderState, ParamsContext>,
-  serviceJSON: ServiceJSON
+  serviceJSON: ServiceJSON,
+  globalRateLimitBucketPerMinute?: TokenBucket
```
victorges, line 95: **nit:** Something short like `globalLimiter` should be enough. We already have all the explicitness in the field name in `service.json`  which is the public interface so there we want to have things self-descriptive, but here we can use some more objective names for the vars.

This applies to other places with this long var name as well, but it is completely optional in all of them as well.

src/service/worker/index.ts
```diff
@@ -211,6 +218,8 @@ export const startWorker = (serviceJSON: ServiceJSON) => {
   const app = new Koa()
   app.proxy = true
   app
+    .use(error)
+    .use(concurrentRateLimiter(serviceJSON?.globalRateLimit?.concurrent))
```
victorges, line 222: I may have suggested you to put this middleware here, but looking closer now I think it should go after all the observability middlewares below. So put it between the `addMetricsLoggerMiddleware` and the `compress` middlewares instead (`error` middleware should still be the first one though).

src/service/worker/index.ts
```diff
@@ -224,9 +233,10 @@ export const startWorker = (serviceJSON: ServiceJSON) => {
     scaleClientCaches(serviceJSON.workers, clients.options)
   }
 
-  const appHttpHandlers = createAppHttpHandlers(service, serviceJSON)
+  const globalRateLimitBucketPerMinute: TokenBucket | undefined = createTokenBucket(serviceJSON?.globalRateLimit?.perMinute)
+  const appHttpHandlers = createAppHttpHandlers(service, serviceJSON, globalRateLimitBucketPerMinute)
+  const appEventHandlers = createAppEventHandlers(service, serviceJSON, globalRateLimitBucketPerMinute)
   const appGraphQLHandlers = createAppGraphQLHandler(service, serviceJSON)
```
victorges, line 239: Aren't you forgetting to send the global limiter to this graphql handler creator as well?

This is an error that makes me propose a nit: maybe these functions should not receive a `globalRateLimitBucketPerMinute?: TokenBucket`, which would allow them to _not_ receive the argument at all, but instead they should receive a `globalRateLimitBucketPerMinute: TokenBucket | undefined` so that the transpiler will complain if we forget about it like hre.

src/service/worker/runtime/http/middlewares/error.ts
```diff
@@ -86,11 +86,7 @@ const logAndRemoveSensitiveData = <
   }
 }
 
-export async function error<
-  T extends IOClients,
-  U extends RecorderState,
-  V extends ParamsContext
-> (ctx: ServiceContext<T, U, V>, next: () => Promise<void>) {
+export async function error (ctx: ServiceContext, next: () => Promise<void>) {
```
victorges, line 89: If I wasn't there, I would ask what this change was for 😆 

FTR: This broke the typings when we used the middleware in the beginning of the pipeline. We could simply remove the type annotations to fix it and let TS do the magic.

src/service/worker/runtime/http/middlewares/rateLimit.ts
```diff
@@ -0,0 +1,45 @@
+import TokenBucket from 'tokenbucket'
+import { TooManyRequestsError } from '../../../../../errors'
+import { ServiceContext } from '../../typings'
+import { createTokenBucket } from '../../utils/tokenBucket'
+
+const responseMessageConcurrent = 'Rate Exceeded: Too many requests in execution'
+const responseMessagePerMinute = 'Rate Exceeded: Too many requests'
```
victorges, line 7: **nit:** WDYT of adding the `per minute` in the end?

src/typings/tokenbucket.d.ts
```diff
@@ -0,0 +1,24 @@
...
+  import Promise = require('bluebird')
+  import redis = require('redis')
+  declare class TokenBucket {
+      public lastFill: any
+  
+      public tokensToAddPerInterval: any
+  
+      public tokensLeft: any
+  
+      constructor(config?: any);
```
victorges, line 11: Would be nice to have real typings for this config as well. Doesn't that tool generate it as well?

src/service/worker/runtime/http/middlewares/rateLimit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+import { ServiceContext } from '../../typings'
+import { createTokenBucket } from '../../utils/tokenBucket'
+import { concurrentRateLimiter, perMinuteRateLimiter } from './rateLimit'
+
+const nopCtx: ServiceContext = {} as ServiceContext
+const nopNext = async () => { return }
+const mutexNext = (mutex: Semaphore) => () => mutex.wait()
+const throwNext = () => { throw new ForbiddenError('Default message')}
+
+async function execRequests(requestsAmount: number, middleware: (ctx: ServiceContext, next: () => Promise<void>) => Promise<void>) {
```
victorges, line 13: **nit:** For easier digesting of the whole code here, you could create a "typedef" for this function type to avoid repeating it everywhere you need to pass it around. i.e.:
```suggestion
type Middleware = (ctx: ServiceContext, next: () => Promise<void>) => Promise<void>

async function execRequests(requestsAmount: number, middleware: Middleware) {
```

Then just use `Middleware` everywhere else below.

src/service/worker/runtime/http/middlewares/rateLimit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+  for (let i = 0; i < requestsAmount; i++) {
+    await middleware(nopCtx, nopNext)
+  }
+}
+
+function startRequest(middleware: (ctx: ServiceContext, next: () => Promise<any>) => Promise<void>, customNext?: () => Promise<void>) {
+  const mutex: Semaphore = new Semaphore(0)
+  const middlewarePromise = middleware(nopCtx, customNext ? customNext : mutexNext(mutex))
```
victorges, line 21: **nit:** 
```suggestion
  const middlewarePromise = middleware(nopCtx, customNext ?? mutexNext(mutex))
```

src/service/worker/runtime/http/middlewares/rateLimit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+  const mutexArr: Semaphore[] = []
+  const promisesArr: Array<Promise<void>> = []
+  
+  for (let i = 0; i < requestsAmount; i++) {
+    const request = startRequest(middleware, customNext)
+    mutexArr.push(request.mutex)
+    promisesArr.push(request.middlewarePromise)
+  }
+  
+  return { mutexArr, promisesArr }
```
victorges, line 39: **nitnit:** How about some functional? 🤓 
```suggestion
  const results = Array(requestsAmount).map(_ => startRequest(middleware, customNext))
  return { 
    mutexArr: results.map(r => r.mutex), 
    promisesArr: results.map(r => r.middlewarePromise),
  }
```

It's a bit less efficient but it's just tests, we can have some fun 🤪
And you can probably write it better, having more JS knowledge than me, or maybe using `ramda` 🌚  (but don't overdo it, just playing around hehe)

src/service/worker/runtime/http/middlewares/rateLimit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  test('Test global per minute middleware with short limit', async () => {
+    const globalRateLimitBucketPerMinute: TokenBucket | undefined = createTokenBucket(2 * 500)
+    const perMinuteRateLimiterMiddlewareRoute = perMinuteRateLimiter(2 * 1000, globalRateLimitBucketPerMinute)
+    const perMinuteRateLimiterMiddlewareEvent = perMinuteRateLimiter(2 * 1000, globalRateLimitBucketPerMinute)
+
+    await expect(
+      Promise.all([
+        execRequests(600, perMinuteRateLimiterMiddlewareRoute),
+        execRequests(600, perMinuteRateLimiterMiddlewareEvent),
```
victorges, line 66: How about `300` in each instead, to make sure that by hitting only the global limit it throws? (not that it would currently trigger the local limits, but the sum would, in case there was a bug somehow combining them)

src/service/worker/runtime/http/middlewares/rateLimit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+})
+
+describe('Rate limit concurrent', () => {
+  test('Test concurrent middleware not throw 429', async () => {
+    const concurrentLimiterMiddleware = concurrentRateLimiter(5)
+
+    const firstHalfRequests = getRequestsArray(5, concurrentLimiterMiddleware)
+    for (let i = 0; i < 5; i++) {
+      firstHalfRequests.mutexArr[i].signal()
+    }
```
victorges, line 94: **nit:** 
```suggestion
    firstHalfRequests.mutexArr.forEach(m => m.signal())
```

Otherwise, I think it's also generally better to do a `for (var m of mutexArr)` loop instead of manually managing the loop index.

Also applies (optionally) to similar code below.

src/service/worker/runtime/http/middlewares/rateLimit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+      Promise.all(firstPart.promisesArr)
+    ).resolves.not.toThrowError(TooManyRequestsError)
+
+    for (let i = 0; i < 5; i++) {
+      secondPart.mutexArr[i].signal()
+    }
+
+    await expect(
+      Promise.all(secondPart.promisesArr)
+    ).rejects.toThrowError(TooManyRequestsError)
```
victorges, line 129: How about checking this before even signaling any mutex at all? I think that on line 114 these should all be rejected already.

src/service/worker/runtime/http/middlewares/rateLimit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    ).rejects.toThrowError(ForbiddenError)
+
+    const secondHalfRequests = getRequestsArray(5, concurrentLimiterMiddleware)
+    for (let i = 0; i < 5; i++) {
+      secondHalfRequests.mutexArr[i].signal()
+    }
+
+    await expect(
+      Promise.all(secondHalfRequests.promisesArr)
+    ).resolves.not.toThrowError(TooManyRequestsError)
```
victorges, line 160: Is there a way to expect it not to throw any error at all? i.e. if these threw `ForbiddenError`, would this test pass?

src/service/worker/runtime/http/middlewares/rateLimit.test.ts
```diff
@@ -0,0 +1,162 @@
+import Semaphore from 'semaphore-async-await'
```
victorges, line 1: Really great work with these tests! 

All comments in this file are completely optional `nits`. They would already be in a regular code, but I think it is even less necessary to spend much time refactoring tests code for the best possible outcome (unless it's like a test framework, like the one from apps' `api_test.go`). You also already spent a lot of time on these, so feel free to just ignore everything I suggested here if you prefer 😄

victorges (approved):
I know the diversions in the feature/implementation took you quite some time as the task grew in complexity, but I think the result was really neat! Kudos for the great work here.

(LGTM!)

src/service/worker/runtime/http/middlewares/rateLimit.test.ts
VerasThiago: If you expect to `throw` any error that is not `TooManyRequests` you need to change to 

```
await expect(
    Promise.all(secondHalfRequests.promisesArr)
).rejects.not.toThrowError(TooManyRequestsError)
```

src/service/worker/runtime/http/middlewares/rateLimit.ts
```diff
@@ -3,19 +3,19 @@ import { TooManyRequestsError } from '../../../../../errors'
 import { ServiceContext } from '../../typings'
 import { createTokenBucket } from '../../utils/tokenBucket'
 
-const responseMessageConcurrent = 'Rate Exceeded: Too many requests in execution'
-const responseMessagePerMinute = 'Rate Exceeded: Too many requests'
+const responseMessageConcurrent = 'Rate Exceeded: Too many requests in execution per minute'
```
victorges, line 6: The requests in execution are not per minute, so you should remove the `per minute` suffix from this one.

src/typings/tokenbucket.d.ts
```diff
@@ -1,22 +1,43 @@
 declare module 'tokenbucket' {
   import Promise = require('bluebird')
   import redis = require('redis')
+
+  declare interface TokenBucketOptions {
```
victorges, line 5: Nice! :)

src/service/worker/runtime/http/middlewares/rateLimit.test.ts
```diff
@@ -9,16 +9,17 @@ const nopCtx: ServiceContext = {} as ServiceContext
...
-  for (let i = 0; i < requestsAmount; i++) {
-    await middleware(nopCtx, nopNext)
-  }
+async function execRequests(requestsAmount: number, middleware: Middleware) {
+  await Promise.all(
+    new Array(requestsAmount).fill(null).map(_ => middleware(nopCtx, nopNext))
+  )
 }
 
 function startRequest(middleware: (ctx: ServiceContext, next: () => Promise<any>) => Promise<void>, customNext?: () => Promise<void>) {
```
victorges, line 20: You can use the `Middleware` type in all these similar declarations as well :)

src/service/worker/runtime/http/index.ts
```diff
@@ -56,7 +56,7 @@ export const createPublicHttpRoute = <T extends IOClients, U extends RecorderSta
   serviceHandler: RouteHandler<T, U, V> | Array<RouteHandler<T, U, V>>,
   serviceRoute: ServiceRoute,
   routeId: string,
-  globalRateLimitBucketPerMinute?: TokenBucket
+  globalLimiter?: TokenBucket
```
victorges, line 59: Should we use `globalLimiter: TokenBucket | undefined` instead of `globalLimiter?: TokenBucket` in all these handler function arguments as well?

victorges (approved):
Clean LGTM! Nice work 👌

athoscouto (approved):
In Victor we trust.

PR #449, opened 2021-05-11
Bypass metrics handling for internet requests

brunoabreu:

Stop exposing the `/metrics` endpoint used by Prometheus.

Internal metadata leaking.

It is working here: https://brunoh2--b2bstore.myvtex.com/metrics
The leak can be seen anywhere else, like: https://storetheme.vtex.com/metrics

#### Types of changes

### DIFF
--- a/src/service/tracing/tracingMiddlewares.ts
+++ b/src/service/tracing/tracingMiddlewares.ts
@@ -18,7 +18,7 @@ import {
   RequestsMetricLabels,
 } from './metrics/instruments'
 
-const PATHS_BLACKLISTED_FOR_TRACING = ['/metrics', '/_status', '/healthcheck']
+const PATHS_BLACKLISTED_FOR_TRACING = ['/_status', '/healthcheck']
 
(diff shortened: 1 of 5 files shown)


src/service/tracing/tracingMiddlewares.ts
```diff
@@ -18,7 +18,7 @@ import {
   RequestsMetricLabels,
 } from './metrics/instruments'
 
-const PATHS_BLACKLISTED_FOR_TRACING = ['/metrics', '/_status', '/healthcheck']
+const PATHS_BLACKLISTED_FOR_TRACING = ['/_status', '/healthcheck']
```
Jeymisson, line 21: I think we still want to keep it blocked. We don't want to trace prometheus requests I guess.

Jeymisson (approved):
lgtm

src/service/tracing/tracingMiddlewares.ts
brunoabreu: It won't be traced. Prometheus middleware is now running before this one.

PR #455, opened 2021-07-13
NAD-606 - Add `@metric` directive

marcelovicentegc:

To add a new directive to easily log data from GraphQL operations to Splunk.

Example usage:

```graphql
type Query {
    someOperation: String! @metric
    anotherOperation: String! @metric(name: "another-operation")
    metriclessOperation: String!
}
```

Currently, only GraphQL errors are logged to Splunk, but what about the success cases? The goal of this PR is to easily allow successful and unsuccessful operations to be logged to Splunk. This gives us the ability to analyze operation-specific success/error ratios without too much hassle.

The motivation behind this feature is the need to **measure the availability** for the `navigation` query from the `vtex.admin-graphql` application.

Link this application and the [`feat/telemetry` branch from `vtex.admin-graphql`](https://github.com/vtex/admin-graphql/tree/feat/telemetry) on the workspace you want, [toggle this condition to true on the node-vtex-api](https://github.com/vtex/node-vtex-api/blob/master/src/service/worker/runtime/statusTrack.ts#L28) and execute the following query on your workspace: 

```graphql
query Navigation {
  navigation {
      section
  }
}
```

Wait a minute and run this query on Splunk:

```bash
index=io_vtex_logs app IN (vtex.admin-graphql@2*) production=false account={your account} workspace={your workspace} status.name="graphql-metric-*"
```

You should then see the status of the operation you've just executed logged there.

Or, if you don't feel like going through this (I have already):
- [To see a sample of what it should look like once it goes live, click here.
](https://splunk72.vtex.com/en-US/app/vtex_io_apps/search?q=search%20index%3Dio_vtex_logs%20app%20IN%20(vtex.admin-graphql%402*)%20production%3Dfalse%20account%3D%22teamadmin%22%20workspace%3D%22navmetrics%22%20status.name%3D%22graphql-metric-*%22&earliest=1626357540&latest=1626357840&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=&sid=1626358635.3075722_5D3C75C2-A8BC-4E6E-8C2F-95794CD3FA6F)

#### Types of changes

### DIFF
--- a/src/service/worker/runtime/graphql/schema/schemaDirectives/Telemetry.ts
+++ b/src/service/worker/runtime/graphql/schema/schemaDirectives/Telemetry.ts
@@ -1,8 +1,6 @@
 import { defaultFieldResolver, GraphQLField, print } from 'graphql'
 import { SchemaDirectiveVisitor } from 'graphql-tools'
-import { cleanError } from '../../../../../../utils'
-import { Logger, LogLevel } from '../../../../../logger/logger'
-import { logger as globalLogger } from '../../../../listeners'
+import { APP } from '../../../../../..'
 import { GraphQLServiceContext } from '../../typings'
 
 export class Telemetry extends SchemaDirectiveVisitor {
@@ -12,12 +10,12 @@ export class Telemetry extends SchemaDirectiveVisitor {
     field.resolve = async (root, args, ctx, info) => {
       let failedToResolve = false
       let result: any = null
-      let timeToResolve = 0
+      let ellapsed: [number, number] = [0, 0]
 
       try {
-        const start = process.hrtime.bigint()
+        const start = process.hrtime()
         result = await resolve(root, args, ctx, info)
-        timeToResolve = Number(process.hrtime.bigint() - start) / 1000000 // Milliseconds
+        ellapsed = process.hrtime(start)
       } catch (error) {
         result = error
         failedToResolve = true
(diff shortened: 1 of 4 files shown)


src/service/worker/runtime/graphql/schema/schemaDirectives/Telemetry.ts
```diff
@@ -0,0 +1,39 @@
...
+
+      ctx.graphql.status = failedToResolve ? 'error' : 'success'
+
+      const payload = {
+        [ctx.graphql.status]: 1,
+      }
+
+      metrics.batch(`graphql-telemetry-${APP.NAME}-${name}`, failedToResolve ? undefined : ellapsed, payload)
+
+      return result
```
brunoabreu, line 32: I think we don't want to be intrusive here. The original behavior was to throw the error. So I would do something like: 

```ts
if (failedToResolve) {
  throw result
}

return result
```

brunoabreu (changes requested):

src/service/worker/runtime/graphql/schema/schemaDirectives/Telemetry.ts
marcelovicentegc: Good catch 😉 !

brunoabreu (approved):

PR #465, opened 2021-08-18
Add Prometheus client

marcelovicentegc:

To add a Prometheus client to make it easy to create metrics to be scraped by a Prometheus instance from IO apps.

There's no straightforward way to creating custom metrics to be sent to Prometheus so far. The goal of PR is to overcome this limitation.

TODO

TODO

#### Types of changes

* [ ] Bug fix (a non-breaking change that fixes an issue)
* [x] New feature (a non-breaking change that adds functionality)


src/service/tracing/metrics/MetricNames.ts
```diff
@@ -69,3 +69,43 @@ export const BETWEEN_SCRAPES_EVENT_LOOP_LAG_PERCENTILES = {
...
+  help: 'LCP',
+  name: 'frontend_lcp_by_sender',
+  buckets: [10, 20, 40, 80, 160, 320, 640, 1280, 2560, 5120],
+  labelNames: ['sender'],
+  type: METRIC_TYPES.HISTOGRAM,
+}
+
+export const FRONTEND_TTFB_BY_SENDER = {
+  help: 'TTFB',
+  name: 'frontend_ttfb_by_sender',
```
esdrasbrz, line 99: The metric name doesn't have a unit, it should be `seconds` for example. You can read more [here](https://prometheus.io/docs/practices/naming/#metric-names)

src/service/tracing/metrics/MetricNames.ts
```diff
@@ -69,3 +69,43 @@ export const BETWEEN_SCRAPES_EVENT_LOOP_LAG_PERCENTILES = {
...
+  buckets: [10, 20, 40, 80, 160, 320, 640, 1280, 2560, 5120],
+  labelNames: ['sender'],
+  type: METRIC_TYPES.HISTOGRAM,
+}
+
+export const FRONTEND_FID_BY_SENDER = {
+  help: 'TTFB',
+  name: 'frontend_fid_by_sender',
+  buckets: [10, 20, 40, 80, 160, 320, 640, 1280, 2560, 5120],
```
esdrasbrz, line 108: I'm not sure if the same buckets will fit in all metrics. It may vary based on the metric unit and how granular you want on each metric range.

esdrasbrz (changes requested):

src/service/tracing/metrics/MetricNames.ts
marcelovicentegc: Absolutely not! I'm doing some research right now to see if the web vitals provide any directions on the expected thresholds for each metric

marcelovicentegc:
Hi, @esdrasbrz! I moved the FrontendMetrics client to https://github.com/vtex/frontend-metrics-gateway. Why? Creating custom metrics will scale easier if we don't keep this responsibility to the node runtime, once its deployment is quite tricky and just by exposing the Prometheus client we give more freedom for other teams to configure their metrics. 

If everything goes as expected, I will share the https://github.com/vtex/frontend-metrics-gateway app as the frontend custom metrics hub, so we can configure the metrics for all frontend applications there. Node services, on the other hand, can simply import the Prometheus client from the node runtime and set their custom metrics on the app itself, instead of on a "hub". Let me know what you think!

felippenardi (approved):

esdrasbrz (approved):
LGTM

src/clients/index.ts
```diff
@@ -5,3 +5,6 @@ export * from './janus'
 export * from './IOClient'
 export * from './IOClients'
 export * from './IOGraphQLClient'
+
+import promclient from 'prom-client'
```
gris, line 9: Does this work for backend also?

marcelovicentegc: Yes! Absolutely! The frontend-metrics-gateway is actually a Node service: https://github.com/vtex/frontend-metrics-gateway

marcelovicentegc: You'll be able to define custom metrics on any node service simply by importing this client.

Example from the frontend-metrics-gateway:
```typescript
import { promclient } from '@vtex/api'

import { CLSBySender, FIDBySender, LCPBySender } from './metrics'

export const instruments = {
  admin: {
    createCLSBySenderInstrument: () =>
      new promclient.Histogram(CLSBySender('admin')),
    createFIDBySenderInstrument: () =>
      new promclient.Histogram(FIDBySender('admin')),
    createLCPBySenderInstrument: () =>
      new promclient.Histogram(LCPBySender('admin')),
  },
}
```

gris: I'm not sure what is the purpose of this PR in this case. It seems to me that `prom-client` is an npm package. and what this contribution is doing is just to import this package and then export it right away. Wouldn't it be better if we just leave the users of the package to import it in their own apps?

marcelovicentegc: Yep! The `prom-client` is an npm package and it's currently being used on the Node runtime to export metrics for all node apps to Prometheus, [as it's seen here](https://github.com/vtex/node-vtex-api/blob/5fe23f32bfcf33fee76a7184c0715c5fa59d7e42/src/service/worker/runtime/builtIn/middlewares.ts#L25-#L45). What this PR does is to leverage the existing client from the Node runtime to export metrics to Prometheus. This way, we don't need to create new Prometheus clients for every application that needs custom metrics and scrape those. [We discussed this briefly on this thread, and the related one linked right on the first line 😄 ](https://vtex.slack.com/archives/C2WSEU9QQ/p1629395427059900).

gris: Oh ok!

gris (approved):

PR #477, opened 2021-09-24
Fix session headers

throwaway4312398:

Test what random accounts can do.


throwaway4312398:
Random accounts can create and close PRs on public repos.

PR #493, opened 2022-02-25
Fix / Remove only cookie fields when cleaning logs

valentino-amadeus:
#### Types of changes

### DIFF
--- a/src/utils/json.ts
+++ b/src/utils/json.ts
@@ -1,0 +1,17 @@
+export function cleanJson(json: {[k: string]: any}, targetFields: string[]) {
+    for (const key of json.keys()) {
+        let deleted = false
+        for (const field of targetFields) {
+            if (key === field) {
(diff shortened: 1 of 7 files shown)


src/utils/json.ts
```diff
@@ -0,0 +1,17 @@
...
+            }
+        }
+
+        if (!deleted && typeof json[key] === 'object') {
+            json[key] = cleanJson(json[key], targetFields)
+        }
+    }
+
+    return json
+}
```
tergol1, line 17: Isn't there a lib to do that?

tergol1 (approved):

src/utils/json.ts
valentino-amadeus: We could `stringify -> parse`, but that sounds more expensive :not-sure:
I haven't found any other option.

PR #494, opened 2022-02-25
[1.x] Fix / Remove only cookie fields from logs

valentino-amadeus:
On top of https://github.com/vtex/node-vtex-api/pull/492.

#### Types of changes

### DIFF
--- a/src/utils/json.ts
+++ b/src/utils/json.ts
@@ -1,0 +1,17 @@
+export function cleanJson(json: {[k: string]: any}, targetFields: string[]) {
+    for (const key of json.keys()) {
+        let deleted = false
+        for (const field of targetFields) {
+            if (key === field) {
+                delete json[key]
+                deleted = true
+            }
+        }
+
+        if (!deleted && typeof json[key] === 'object') {
+            json[key] = cleanJson(json[key], targetFields)
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


src/utils/json.ts
```diff
@@ -0,0 +1,17 @@
...
+    for (const key of Object.keys(json)) {
+        let deleted = false
+        for (const field of targetFields) {
+            if (key === field) {
+                delete json[key]
+                deleted = true
+            }
+        }
+
+        if (!deleted && typeof json[key] === 'object') {
```
brunoabreu, line 11: `typeof null` is `object` to javascript..

so you will enter the recursion and will 💥  when doing `Object.keys(null)`

 ```
 > Object.keys(null)
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
 ```

brunoabreu (changes requested):

src/utils/json.ts
valentino-amadeus: [Done](https://github.com/vtex/node-vtex-api/pull/494/commits/e33e1aed654e4672f8020a308bc360d1eeb0e92a). Thanks!

brunoabreu (approved):

PR #495, opened 2022-02-25
[3.x] Fix / Only remove cookie fields when removing sensitive data from logs

valentino-amadeus:
#### Types of changes


src/utils/log.ts
```diff
@@ -1,14 +1,7 @@
...
+const SENSITIVE_FIELDS = ['cookie', 'Cookie', 'vtexIdclientautcookie', 'error']
 
-    SECOND_LEVEL_SENSITIVE_FIELDS.forEach(field => {
-        if (field[0] in log) {
-            delete log[field[0]][field[1]]
-        }
-    })
-}
\ No newline at end of file
+export const cleanLog = (log: any) => {
```
gris, line 5: Why do we have this mismatch between what is expected between `log` and `json` in the function `cleanJson`?

valentino-amadeus: We don't anymore :jajaja:. Thanks!

src/utils/json.ts
```diff
@@ -0,0 +1,17 @@
...
+    for (const key of Object.keys(json)) {
+        let deleted = false
+        for (const field of targetFields) {
+            if (key === field) {
+                delete json[key]
+                deleted = true
+            }
+        }
+
+        if (!deleted && typeof json[key] === 'object') {
```
brunoabreu, line 11: Same issue mentioned [here](https://github.com/vtex/node-vtex-api/pull/494#discussion_r821669949)

brunoabreu (changes requested):

src/utils/json.ts
valentino-amadeus: [Done](https://github.com/vtex/node-vtex-api/pull/495/commits/d4df050618492c97991da4f78c5842329df1d153). Thanks!

brunoabreu (approved):

PR #500, opened 2022-03-14
Fix / Sensitive Fields

valentino-amadeus:
#### Types of changes

### DIFF
--- a/src/utils/log.ts
+++ b/src/utils/log.ts
@@ -1,6 +1,6 @@
 import { cleanJson } from './json'
 
-const SENSITIVE_FIELDS = ['authorization', 'cookie', 'Cookie', 'vtexidclientautcookie', 'vtexIdclientautcookie', 'error']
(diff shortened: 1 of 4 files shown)


src/utils/log.ts
```diff
@@ -1,6 +1,6 @@
 import { cleanJson } from './json'
 
-const SENSITIVE_FIELDS = ['cookie', 'Cookie', 'vtexIdclientautcookie', 'error']
+const SENSITIVE_FIELDS = ['authorization', 'cookie', 'Cookie', 'vtexidclientautcookie', 'vtexIdclientautcookie', 'error']
```
gris, line 3: Aren’t we testing the lower case version of every field? Why we need to check for vtexidclientautcookie 2 times?

valentino-amadeus: [Here it is](https://github.com/vtex/node-vtex-api/pull/500/commits/1c4f31b748931bc8b1fb9cc7e63233dc151a2755). Thank you!

gris (approved):

PR #513, opened 2022-11-29
Make the computed hash that is appended to request behave determinist…

MMartyn:
I've observed cache hit ratios that suggested that there may be an issue with the keys used for caching. Upon digging further into it I noted that the cache key for a page that was unchanged was computing different hashes for the calls to render services. Those difference were the order in which items appeared in the json.

By applying a consistent order to the json which is used to generate a hash we can guarantee that the hash will be computed to the same value each time. This will improve efficiency and performance.

I was able to observe consistent hashing by `yarn link`ing the api into one of the render services and then `yarn link`ing that.

Example of the inconsistencies without applying a sort:
```
settings: {
  'vtex.google-search-console':(},
  'vtex.google-tag-manager':(},
  'vtex.google-shopping': (}
}
```
vs
```
settings: {
  'vtex.google-search-console':(},
  'vtex.google-shopping': (},
  'vtex.google-tag-manager':(}
}
```
#### Types of changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@vtex/api",
-  "version": "6.45.13-beta.3",
+  "version": "6.45.13-beta.6",
   "description": "VTEX I/O API client",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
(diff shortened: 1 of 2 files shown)


Jeymisson (approved):
Great catch!! 

lgtm

PR #534, opened 2023-07-07
XTNSNS-391 Bump graphql-upload to v13.0.0

mairatma:

Upgrade the version of graphql-upload.

We're using a very outdated version of graphql-upload (v8) which isn't compatible with node 16. This causes graphql apps such as pages-graphql to fail for any mutation that uses uploads.

Version v13 works correctly in node 16, unlocking us to upgrade node on pages-graphql and other graphql apps that use upload features.

I managed to manually test by running pages-graphql locally on no 16, with this version of @vtex/api linked to it.

For a final test within IO's environment we need to publish a new version of @vtex/api so that we can use it to test a link of pages-graphql using the 7.x node builder. Once we have the published version I'll do that final test.

**Before**
Screenshot 2023-07-07 at 10 54 34

**After**
Screenshot 2023-07-07 at 10 55 16

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [6.45.19] - 2023-07-12
+### Changed
+- Bump `graphql-upload` version
+
+## [6.45.18] - 2023-07-07
 ### Fixed
 - Metrics to prometheus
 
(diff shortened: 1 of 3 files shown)


arturpimentel (approved):

filafb (commented):
Is it backward compatible with node 12? Currently we are keeping the same node SDK for both version.

mairatma:
> Is it backward compatible with node 12? Currently we are keeping the same node SDK for both version.

Yes, it is! 🙏

mairatma:
> Is it backward compatible with node 12? Currently we are keeping the same node SDK for both version.

I've just triple checked to be sure and it's confirmed: this version works correctly in both node 12 and node 16. 

Regardless of that it would be great to test this within IO's environment when we publish a version of this package to NPM, before applying to service-runtime-node. 

Can I merge this and publish a new patch version now to test? Once I confirm it's working I'll let you guys know so we can move forward with the service-runtime-node change as well.

PR #536, opened 2023-07-07
Remove `prometheusLoggerMiddleware` async typing

arturpimentel:
Last PR had a build problem because of a misplaced `async` directive. Removing that solves the issue

To test:
Run `yarn build`. It should execute without errors


eduardocesb (approved):
LGTM ✅

PR #542, opened 2023-09-01
Fix shouldTrace decision based on sampling decision from edge

Jeymisson:
This is complimentary to the previous PR: https://github.com/vtex/node-vtex-api/pull/541

Jaeger client `tracer.extract` will always return a rootSpan context, the Sampling decision is inside this context, as we can see in it's code implementation, https://github.com/jaegertracing/jaeger-client-node/blob/29365a4475285423490813187ea48136b5135167/src/span_context.js#L186.

Example of tagged trace through header: `jaeger-debug-id: trace-test-sampling`
Screenshot 2023-09-07 at 18 14 04

#### Types of changes

### DIFF
--- a/src/service/tracing/tracingMiddlewares.ts
+++ b/src/service/tracing/tracingMiddlewares.ts
@@ -31,6 +31,7 @@ export const addTracingMiddleware = (tracer: Tracer) => {
     const start = process.hrtime()
     concurrentRequests.inc(1)
     const rootSpan = tracer.extract(FORMAT_HTTP_HEADERS, ctx.request.headers) as undefined | SpanContext
+    ctx.tracing = { tracer, currentSpan: undefined}
 
     if (!shouldTrace(ctx, rootSpan)) {
       await next()
@@ -93,7 +94,7 @@ export const addTracingMiddleware = (tracer: Tracer) => {
 
         currentSpan.log(cloneAndSanitizeHeaders(ctx.request.headers, 'req.headers.'))
         currentSpan.log(cloneAndSanitizeHeaders(ctx.response.headers, 'res.headers.'))
-        ctx.set(TRACE_ID_HEADER, traceInfo.traceId)
+        ctx.set(TRACE_ID_HEADER, traceInfo.traceId!)
       }
 
       const onResFinished = () => {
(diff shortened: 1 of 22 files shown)


src/service/tracing/tracingMiddlewares.ts
```diff
@@ -152,6 +152,6 @@ export const traceUserLandRemainingPipelineMiddleware = () => {
 }
 function shouldTrace(ctx: ServiceContext, rootSpan: SpanContext | undefined) {
   // Should trace if path isnt blacklisted and tracing decision came from the edge
-  return !PATHS_BLACKLISTED_FOR_TRACING.includes(ctx.request.path) && rootSpan != null
+  return !PATHS_BLACKLISTED_FOR_TRACING.includes(ctx.request.path) && ((rootSpan as any).isSampled?.() ?? false)
```
Jeymisson, line 155: This is a tiny typehack, SpanContext type is defined by opentracing, but the actual full Jaeger type has a method called isSampled, that's why we cast the SpanContext to any before accessing the isSampled method. This method could be seen here: https://github.com/jaegertracing/jaeger-client-node/blob/29365a4475285423490813187ea48136b5135167/src/span_context.js#L186.

Jeymisson: It makes sense, that's kinda out of Opentracing pattern. Indeed I was expecting the `extract` method to solve all we needed. With no tracing header this method could return null, that's not the case, and why we need to use the `isSampled` one.

From Jaeger's code, `isSampled` means whether or not this span context was sampled. When it is sampled means that this context is being traced.

arturpimentel: Hmmm... so it works like an `isActiveTrace` or something like that. As long as we put a comment above that line explaining this stuff, it's all good to me

arturpimentel: Or just update the comment that already exists :P

Jeymisson: Kinda, to me `isActiveTrace` would be something more wide, independent of a request being sampled or not, i.e: "The tracing is always active, sometimes it samples a request, sometimes not".

When a spanContext `isSampled` this means the initial decision was to trace that request, if this means `isActiveTrace` to you then yeah. Added a comment in there.

src/service/tracing/tracingMiddlewares.ts
```diff
@@ -93,7 +94,7 @@ export const addTracingMiddleware = (tracer: Tracer) => {
 
         currentSpan.log(cloneAndSanitizeHeaders(ctx.request.headers, 'req.headers.'))
         currentSpan.log(cloneAndSanitizeHeaders(ctx.response.headers, 'res.headers.'))
-        ctx.set(TRACE_ID_HEADER, traceInfo.traceId)
+        ctx.set(TRACE_ID_HEADER, traceInfo.traceId!)
```
arturpimentel, line 97: Looks like jaeger's types can't be trusted haha :P

arturpimentel (approved):
Looks like we had a lot of nullchecks to do haha. It LGTM if we can reproduce the conditions for the errors we had last time

arturpimentel (approved):

PR #546, opened 2023-10-02
Fix metrics being restricted by tracing restrictions

mairatma:
Stop restricting request metrics according to the tracing sampling logic. Metrics should always be logged.

Some request metrics were not being logged as frequently as desired after some tracing restrictions were added. That's because their code was tightly mixed with the tracing code, and so this happened by mistake.

To solve this, I've split the tracing middleware into two: one just for tracing and another just for those metrics. The tracing check is only done on the first.

I've navigated through [this workspace](https://mairametrics--hmchile.myvtex.com/) that includes the change and:
- Verified that the store works as before, with no errors.
- Verified that the metrics are being logged regardless of tracing being enabled (checked via console.logs).
- Verified that the traces are not being logged when disabled (checked via console.log and the jaeger-debug-id header).
- Verified that the traces are being logged when enabled (checked via console.log and the jaeger-debug-id header).

#### Types of changes

### DIFF
--- a/src/service/metrics/requestMetricsMiddleware.ts
+++ b/src/service/metrics/requestMetricsMiddleware.ts
@@ -1,0 +1,70 @@
+import { finished as onStreamFinished } from 'stream'
+import { hrToMillisFloat } from '../../utils'
+import { ServiceContext } from '../worker/runtime/typings'
+import {
+  createConcurrentRequestsInstrument,
+  createRequestsResponseSizesInstrument,
+  createRequestsTimingsInstrument,
+  createTotalAbortedRequestsInstrument,
+  createTotalRequestsInstrument,
+  RequestsMetricLabels,
+} from '../tracing/metrics/instruments'
+
+
+export const addRequestMetricsMiddleware = () => {
+  const concurrentRequests = createConcurrentRequestsInstrument()
+  const requestTimings = createRequestsTimingsInstrument()
+  const totalRequests = createTotalRequestsInstrument()
+  const responseSizes = createRequestsResponseSizesInstrument()
+  const abortedRequests = createTotalAbortedRequestsInstrument()
+
+  return async function addRequestMetrics(ctx: ServiceContext, next: () => Promise<void>) {
+    const start = process.hrtime()
+    concurrentRequests.inc(1)
+
+    ctx.req.once('aborted', () =>
+      abortedRequests.inc({ [RequestsMetricLabels.REQUEST_HANDLER]: ctx.requestHandlerName }, 1)
+    )
+
+    let responseClosed = false
+    ctx.res.once('close', () => (responseClosed = true))
+
+    try {
+      await next()
+    } finally {
+      const responseLength = ctx.response.length
+      if (responseLength) {
+        responseSizes.observe(
+          { [RequestsMetricLabels.REQUEST_HANDLER]: ctx.requestHandlerName },
+          responseLength
+        )
+      }
+
+      totalRequests.inc(
+        {
+          [RequestsMetricLabels.REQUEST_HANDLER]: ctx.requestHandlerName,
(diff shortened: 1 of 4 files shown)


src/service/tracing/tracingMiddlewares.ts
```diff
@@ -120,7 +77,8 @@ export const addTracingMiddleware = (tracer: Tracer) => {
 
 export const nameSpanOperationMiddleware = (operationType: string, operationName: string) => {
   return function nameSpanOperation(ctx: ServiceContext, next: () => Promise<void>) {
-    ctx.tracing?.currentSpan?.setOperationName(`${operationType}:${operationName}`)
+    ctx.requestHandlerName = `${operationType}:${operationName}`
```
mairatma, line 80: I've added this new field to the context because the metrics were using it to include extra data, but this was only being added when tracing was enabled. Now we can access this name regardless of the tracing.

src/service/tracing/tracingMiddlewares.ts
```diff
@@ -7,35 +7,19 @@ import { RuntimeLogFields } from '../../tracing/LogFields'
...
 
   return async function addTracing(ctx: ServiceContext, next: () => Promise<void>) {
-    const start = process.hrtime()
-    concurrentRequests.inc(1)
     const rootSpan = tracer.extract(FORMAT_HTTP_HEADERS, ctx.request.headers) as undefined | SpanContext
     ctx.tracing = { tracer, currentSpan: undefined}
 
     if (!shouldTrace(ctx, rootSpan)) {
       await next()
-      concurrentRequests.dec(1)
```
danzanzini, line 38: Nice catch 👏

danzanzini (approved):
LGTM :shipit: 

Do this explain the strage behavior we saw on the metric?

mairatma:
> Do this explain the strage behavior we saw on the metric?
Maybe, we have to deploy this to see how the metric will change. We'll be able to see of those seemingly always increasing numbers of concurrent requests were due to errors skipping to decrement the metric or if they really are increasing.

src/service/metrics/requestMetricsMiddleware.ts
```diff
@@ -0,0 +1,70 @@
...
+  return async function addRequestMetrics(ctx: ServiceContext, next: () => Promise<void>) {
+    const start = process.hrtime()
+    concurrentRequests.inc(1)
+
+    ctx.req.once('aborted', () =>
+      abortedRequests.inc({ [RequestsMetricLabels.REQUEST_HANDLER]: ctx.requestHandlerName }, 1)
+    )
+
+    let responseClosed = false
+    ctx.res.once('close', () => (responseClosed = true))
```
filafb, line 30: When the request closes, do we still run the `finally` clause? Instead of rellying on it and in the global variable, could we pass the `onResFinished` to run here? WDYT?

filafb (approved):
The code LGTM. Just added a question that I genuinely don't know the answer 😅

src/service/metrics/requestMetricsMiddleware.ts
mairatma: Nope, this is an event, not related to errors that can be wrapped with try/catch/finally. You need to listen to it to really be sure of the event happening.

I also didn't change the way the code handles this or any other part of the existing logic, to minimize any side effects of this refactor. I just copied/pasted everything that was related to metrics to this file, which includes waiting for the request to be closed to set some of these metrics. You'll notice that the original file (tracingMiddleware.ts) contains this same logic, but that now it only runs the tracing-related code.

filafb: Oh, I see. It's a request event, not a process event. I was afraid it would terminate the whole process, then skipping anything else. But I got the gist, thanks!

PR #549, opened 2023-10-05
Stop updating client cache when revalidated and still expired

mairatma:

This is removing a bunch of unnecessary disk writes, which just waste CPU.

When a cached response is expired, we try sending a request using the cached etag to revalidate it. If this revalidation is successful, we use the cache contents. In that case, we also update the cache again **with the contents it already has**. That's just a waste of CPU.

This is only ever useful if the cache max age has increased, to avoid further revalidations, since we also write the new max age to the disk. But at least for pages-graphql calls it's very frequent for the content to still be expired at this point. 

This PR is making sure that we only ever update the cache if either the content has changed (meaning that we didn't use the cached contents) or the expiration time has changed.

This should free a lot of disk writes from pages-graphql (at least ~60% of disk writes for VBase calls, plus others in the same situation).

I've tested this by vtex linking pages-graphql with @vtex/api locally linked, and adding console.log to show when we skip writing to cache. The skips were done at the right time (after revalidating VBase requests, which are always expired).

I've also tested by looking at Honeycomb traces. On production you'll always see the `local-cache-saved` event for requests to VBase, even for STALE requests. Now you'll only ever see this event for MISS requests to VBase ([example trace with STALE](https://ui.honeycomb.io/vtex/datasets/vtex/result/xGhYKiApJPn/trace/nWPYfWDvwDr?fields[]=c_name&fields[]=c_service.name&fields[]=c_http.cache.disk&span=447dc3b2be8ae28b), [example trace with MISS](https://ui.honeycomb.io/vtex/datasets/vtex/result/p5Unmwgsj6n/trace/oJDszfxobHe?fields[]=c_name&fields[]=c_service.name&fields[]=c_http.cache.disk&span=1649e70a8fae547d)).

#### Types of changes

### DIFF
--- a/src/HttpClient/middlewares/cache.ts
+++ b/src/HttpClient/middlewares/cache.ts
@@ -198,7 +198,16 @@ export const cacheMiddleware = ({ type, storage }: CacheOptions) => {
         ? (data as Buffer).toString(responseEncoding)
         : data
 
-      const expiration = Date.now() + (maxAge - currentAge) * 1000
+      const now = Date.now()
+      const expiration = now + (maxAge - currentAge) * 1000
+
+      const alreadyExpired = expiration <= now
+      const reusingRevalidatedCache = cached && (ctx.response === cached.response)
+      const shouldSkipCacheUpdate = alreadyExpired && reusingRevalidatedCache
+      if (shouldSkipCacheUpdate) {
+        return
+      }
+
       await storage.set(setKey, {
         etag,
         expiration,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- Stop updating client cache when revalidated and still expired
+
 ## [6.45.23] - 2023-10-04
 
 ### Fixed


src/HttpClient/middlewares/cache.ts
```diff
@@ -198,7 +198,16 @@ export const cacheMiddleware = ({ type, storage }: CacheOptions) => {
         ? (data as Buffer).toString(responseEncoding)
         : data
 
-      const expiration = Date.now() + (maxAge - currentAge) * 1000
+      const now = Date.now()
+      const expiration = now + (maxAge - currentAge) * 1000
+
+      const alreadyExpired = expiration <= now
+      const reusingRevalidatedCache = cached && (ctx.response === cached.response)
```
filafb, line 205: Should we use the logic from line 156 here? The `ctx.response === cached.response` will only be true when we enter that branch, won't it?

filafb (commented):
For VBase requests, this implementation makes the remove of `await` from `storage.set` unnecessary. Should we merge the changes on [PR](https://github.com/vtex/node-vtex-api/pull/551) and test it all together? We are trying to create a test env to evaluate the changes before going live with it. WDYT?

mairatma:
> For VBase requests, this implementation makes the remove of `await` from `storage.set` unnecessary. Should we merge the changes on [PR](https://github.com/vtex/node-vtex-api/pull/551) and test it all together? We are trying to create a test env to evaluate the changes before going live with it. WDYT?

Do you mean [this test environment](https://vtex.slack.com/archives/C05PRQTJJ57/p1698062829107089)? I like the idea of having that but it seems unnecessary for this change, since I was able to test it very well by just linking, with console.log + tracing. What's the main purpose of using this new test environment? What kind of changes would require using it?

As for merging with the PR for running the saves on the background, I feel like it's better to deploy them separately. This one is ready and it has some extra logic related to checking expiration. It would already help VBase's case, like you said. It would be better to observe it on its own before deploying writes on the background, since those will affect other clients besides VBase.

src/HttpClient/middlewares/cache.ts
mairatma: I think that this check here is more complete, as it would still work for any future use cases besides the existing 304. I think it's more semantic as well, making it clear that we want to check if we're using the cached contents, regardless of the reason why we used them.

filafb: Cool. It makes sense to me. The only think I don't like is using the `===` for objects in JS. It can be misleading, since we are only comparing the reference, but let's move forward with that.

filafb:
Not necessary that one, but any test env that allow us a `close to production` test. The main idea is to avoid having side effects we cannot map testing only linking the projects and having more data to back up the changes. But since we will be moving this change by itself, I think we are safe.

filafb (approved):
LGTM 🚀

filipewl (approved):
Looks good to me! 

We are only missing an entry to the change log. 📝

src/HttpClient/middlewares/cache.ts
```diff
@@ -198,7 +198,16 @@ export const cacheMiddleware = ({ type, storage }: CacheOptions) => {
         ? (data as Buffer).toString(responseEncoding)
         : data
 
-      const expiration = Date.now() + (maxAge - currentAge) * 1000
+      const now = Date.now()
+      const expiration = now + (maxAge - currentAge) * 1000
+
+      const alreadyExpired = expiration <= now
```
arturpimentel, line 204: I was confused at first here by this math, but it seems to me that given from L202:

`expiration - now = 1000 * (maxAge - currentAge)`

Then `expiration <= now` if, and only if, `maxAge <= currentAge`, which means `alreadyExpired` maybe could be simplified:


```suggestion
      const alreadyExpired = maxAge <= currentAge
```

arturpimentel (commented):
LGTM! I left a minor comment

src/HttpClient/middlewares/cache.ts
arturpimentel: Oh, I see. I did not pay attention to what we had before :grimacing:

In that case I agree your change does not decrease the code clarity, since my eyes were drawn to that whole `const expiration = now + (maxAge - currentAge) * 1000` that was already there, and it doesn't seem like it will disappear haha. Let's move on.

arturpimentel (approved):

PR #551, opened 2023-10-16
Dont await to write cache in disk to respond to client

filafb:
#### Types of changes

### DIFF
--- a/src/HttpClient/middlewares/cache.ts
+++ b/src/HttpClient/middlewares/cache.ts
@@ -199,7 +199,7 @@ export const cacheMiddleware = ({ type, storage }: CacheOptions) => {
         : data
 
       const expiration = Date.now() + (maxAge - currentAge) * 1000
-      await storage.set(setKey, {
+      storage.set(setKey, {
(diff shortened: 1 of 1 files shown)


src/HttpClient/middlewares/cache.ts
```diff
@@ -199,7 +199,7 @@ export const cacheMiddleware = ({ type, storage }: CacheOptions) => {
         : data
 
       const expiration = Date.now() + (maxAge - currentAge) * 1000
-      await storage.set(setKey, {
+      storage.set(setKey, {
```
mairatma, line 202: I know this is still in Draft state, but just a suggestion: I feel like the `span?.log` call on line 210 below should only happen after the write finishes, so that we only log this on success and so that the trace shows the right timestamp for this event.

To do this we'll need to make sure that adding an event to a span after it's closed will work though. It might not be possible. Maybe this event won't make sense within that span anymore actually, since it's an event that happens afterwards, not tied to this span anymore. It'll just be background work now.

PR #554, opened 2023-11-08
Change LRUDiskCache to allow subclasses to override functionality

mairatma:

This changes LRUDiskCache so that a subclass can properly override some of its functionality (such as the has/get/set methods, stats calculations, etc).

We need to build a new disk cache implementation with extra features. To avoid reimplementing everything again, we can instead extend from LRUDiskCache and override some of its methods to achieve what we need. But the way LRUDiskCache was created didn't allow for these extensions, so this PR is changing it to allow better usage by subclasses.

- Build the codebase and verify that there were no errors.
- Create a new class extending from LRUDiskCache and make sure that there are no errors when overriding the methods.

#### Types of changes


src/caches/LRUDiskCache.ts
```diff
@@ -33,11 +36,26 @@ export class LRUDiskCache<V> implements CacheLayer<string, V>{
...
+    this.lruStorage = new LRU<string, LRUData>(lruOptions)
+
+  }
 
+  /**
+   * Builds the data object that will be stored at the LRU memory storage.
+   * Subclasses that need to store more than just the time of death should
+   * override this.
+   */
+  protected buildLRUData(timeOfDeath: number, localCacheOptions?: LocalCacheOptions): LRUData {
```
mairatma, line 48: The data stored in the memory part of the cache was only the timeOfDeath. This new method will allow storing more data there, for any logic that needs to run before deciding to fetch the file from disk.

In our use case this extra data will be the list of fields that we have stored from the original object.

src/caches/LRUDiskCache.ts
```diff
@@ -33,11 +36,26 @@ export class LRUDiskCache<V> implements CacheLayer<string, V>{
...
+    return { timeOfDeath }
   }
 
-  public has = (key: string): boolean => this.lruStorage.has(key)
+  protected getLRU() {
+    return this.lruStorage
+  }
+
+  public has (key: string): boolean {
```
mairatma, line 56: I changed the has/get/set methods syntax because the previous syntax doesn't work properly when a subclass overrides, losing the access to `super`. No changes were made here besides the syntax though.

src/caches/LRUDiskCache.ts
```diff
@@ -85,23 +103,18 @@ export class LRUDiskCache<V> implements CacheLayer<string, V>{
     })
 
     // if it is an outdated file when stale=true
-    if (timeOfDeath < Date.now()) {
+    if (lruData.timeOfDeath < Date.now()) {
```
mairatma, line 106: The `timeOfDeath` is now part of the data retrieved from the memomery cache (lruStorage), instead of being the value itself.

src/caches/LRUDiskCache.ts
```diff
@@ -124,6 +124,19 @@ export class LRUDiskCache<V> implements CacheLayer<string, V>{
...
   }
 
+  /**
+   * Builds the data object that will be stored at the LRU memory storage.
+   * Subclasses that need to store more than just the time of death should
+   * override this.
+   */
+  protected buildLRUData(timeOfDeath: number, localCacheOptions?: LocalCacheOptions): LRUData {
+    return { timeOfDeath }
+  }
```
mairatma, line 134: The data stored in the memory part of the cache was only the timeOfDeath. This new method will allow storing more data there, for any logic that needs to run before deciding to fetch the file from disk.

In our use case this extra data will be the list of fields that we have stored from the original object.

src/caches/LRUDiskCache.ts
```diff
@@ -85,23 +90,18 @@ export class LRUDiskCache<V> implements CacheLayer<string, V>{
...
-    if (maxAge) {
-      timeOfDeath = maxAge + Date.now()
-      this.lruStorage.set(key, timeOfDeath, maxAge)
-    }
-    else {
-      this.lruStorage.set(key, NaN)
-    }
+  public async set (key: string, value: V, maxAge?: number, localCacheOptions?: LocalCacheOptions): Promise<boolean> {
+    const timeOfDeath = maxAge ? maxAge + Date.now() : NaN
+    const lruData = this.buildLRUData(timeOfDeath, localCacheOptions)
```
mairatma, line 103: This just changed the way we build the data to be stored in the memory cache to use the new buildLRUData function.

PR #556, opened 2024-01-29
Increase messages timeout

filafb:

We have seen some timeout errors in the admin coming from `translateWithDependencies` method. We are increasing the timeout to the MessageGraphQL to accommodate the P99 duration time for this method - [here](https://ui.honeycomb.io/vtex/datasets/vtex/result/6GR5x8MjSry). We are also increasing the number of concurrent request to messages, since the app is running below 50% of CPU usage on average.


eduardocesb (approved):
LGTM ✅

arturpimentel (approved):

PR #562, opened 2024-10-03
Delete Metadata client

iagoaraujo:

Delete the Metadata client as it is not used by any app, and the main reason is that there are no policies available at Sphinx to get the access to its endpoints.

Also I've check the logs to make sure that no one is using it (even receiving forbidden, we could have apps with the Metadata client on their codes). I didn't find any app using it.

#### Types of changes


silvadenisaraujo (approved):
LGTM 👍

PR #563, opened 2024-10-14
@auth: Validate token's `account` scope

arturpimentel:

Check whether the token's indicated account is the request's context account.

Until now, the @auth logic allows for cross-account requests as long as the target account contains the user that is identified by the token. This is not usually a problem since the token has to be a valid VTEX ID token and the user mail has to have the necessary access scope in the target account in order for this to work. However, it is a conceptual error (and a security flaw) to allow for cross-account requests in this manner unless the resource is explicitly allowed by a ReBAc rule, which is not the case here (GraphQL fields are not ReBAc-sensitive in VTEX IO).

- Create/Use a user that exists in `vtexgame1`, but in a different account;
- Take an admin token from that user in that account;
Use that token to make the following request:

```shell
curl --location 'https://testauth--vtexgame1.myvtex.com/_v/private/graphql/v1' \
--header "Cookie: ${token}" \
--header 'Content-Type: application/json' \
--data-raw '{"query":"{ \n    availableApps() @context(provider: \"vtex.apps-graphql@3.x\") {    id    vendor    version    name    slug    icon    title    featured    categories    registry  }}","variables":{}}'
```

You should get an 200 response but with an error body like the one below:
```
{
    "data": {
        "availableApps": null
    },
    "errors": [
        {
            "message": "Could not find user specified by VtexIdclientAutCookie.",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "vtex_appsgraphql_3_17_4_availableApps"
            ],
            "extensions": {
                "code": "UNAUTHENTICATED",
                "exception": {
                    "message": "Could not find user specified by VtexIdclientAutCookie.",
                    "name": "AuthenticationError",
                    "extensions": {
                        "code": "UNAUTHENTICATED"
                    },
                    "sensitive": {
                        "stack": "AuthenticationError: Could not find user specified by VtexIdclientAutCookie.\n    at ..."
}
```

#### Types of changes

### DIFF
--- a/src/service/worker/runtime/graphql/schema/schemaDirectives/Auth.ts
+++ b/src/service/worker/runtime/graphql/schema/schemaDirectives/Auth.ts
@@ -1,4 +1,4 @@
-import { ForbiddenError, UserInputError } from 'apollo-server-errors'
+import { AuthenticationError, ForbiddenError, UserInputError } from 'apollo-server-errors'
 import axios from 'axios'
 import { defaultFieldResolver, GraphQLField } from 'graphql'
 import { SchemaDirectiveVisitor } from 'graphql-tools'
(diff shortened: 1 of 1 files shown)


silvadenisaraujo (approved):
LGTM :shipit:

PR #577, opened 2025-04-02
Update logging instrumentation

daniyelnnr:

Update logging instrumentation, adding telemetry and log clients and integrates `node-vtex-api` with the new version of the [diagnostics](https://github.com/vtex/diagnostics) library that supports OpenTelemetry protocols.
It also updates Typescript and Opentelemetry dependencies to enable support for node builder 6.x and IO apps using Node 16.x.

This implementation of `node-vtex-api` is not using the newest version of [diagnostics](https://github.com/vtex/diagnostics), which is VTEX's standard library for instrumentation and observability. This PR provides consistency in telemetry generation and allows it to adopt OTel protocol for logs.

To test it:
- Clone this repository and checkout to `update/logs-instrumentation` branch;
- Install the dependencies and generate the build:
```
  yarn && yarn build
```
- Link the previously generated package:
```
  yarn link
```
- Create an IO app (e.g. service-example)
- Link this package to your app's node_modules
```
  cd ../<your-app>/node && yarn link @vtex/api
```
- Add some logging to your app and monitor new logs being received by the collector through this [dashboard](https://grafana.vtex.com/goto/d6tkoqoNg?orgId=1) in Grafana.

#### Types of changes


src/service/logger/client.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  return initPromise;
+}
+
+async function initializeClient(account: string, workspace: string): Promise<LogClient> {
+  try {
+    const telemetryClient = await getTelemetryClient();
+
+    const logsConfig = Exporters.CreateLogsExporterConfig({
+      endpoint: process.env.OTEL_EXPORTER_OTLP_ENDPOINT || 'http://io-telemetry-http.vtex.systems:3333',
```
daniyelnnr, line 30: I couldn't identify the proper way to handle this address, at first I thought it should be in [constants.ts](https://github.com/vtex/node-vtex-api/blob/master/src/constants.ts) but I'm not sure about that.

src/HttpClient/middlewares/cache.ts
```diff
@@ -109,11 +109,11 @@ export const cacheMiddleware = ({ type, storage, asyncSet }: CacheOptions) => {
 
 
     const cacheReadSpan = createCacheSpan(cacheType, 'read', tracer, span)
-    let cached: void | Cached
+    let cached: void | Cached = undefined
     try {
       const cacheHasWithSegment = await storage.has(keyWithSegment)
       cached = cacheHasWithSegment ? await storage.get(keyWithSegment) : await storage.get(key)
-    } catch (error) {
```
arturpimentel, line 116: These will probably conflict with the changes I made myself in https://github.com/vtex/node-vtex-api/pull/574. I believe I like your implementation better, so I might keep the changes you made here in the future.

src/service/logger/client.ts
arturpimentel: I don't know if we should have a default value for the exporter endpoint. Maybe if we can't initialize it with a environment variable, this instance just doesn't send logs anywhere (or logs them to stdout as a default). I'm afraid of a behaviour where for some reason, if the endpoints are unavailable, this client crashes the application.

arturpimentel (changes requested):

arturpimentel (changes requested):
Would you mind adding a high-level overview of the changes? I had some trouble understanding the logging architecture here. Maybe I'll use copilot for this, but I think it would be beneficial if I could picture how this is organized

src/service/telemetry/client.ts
```diff
@@ -0,0 +1,69 @@
...
+      TelemetryClientSingleton.instance = new TelemetryClientSingleton();
+    }
+    return TelemetryClientSingleton.instance;
+  }
+
+  private async initTelemetryClient(): Promise<TelemetryClient> {
+    try {
+      const telemetryClient = await NewTelemetryClient(
+        'node-vtex-api',
+        APP.ID || 'vtex-app',
```
silvadenisaraujo, line 23: Is there a risk here to don't have the app id available?

src/service/logger/logger.ts
```diff
@@ -86,11 +87,34 @@ export class Logger {
...
+      });
     }
 
-    console.log(JSON.stringify(inflatedLog))
+    if (this.logClient) {
+      try {
+        let logMessage = typeof data === 'string' ? data : JSON.stringify(data)
+        switch (level) {
+          case LogLevel.Debug:
+            this.logClient.debug(logMessage, inflatedLog);
```
silvadenisaraujo, line 100: Looking into the inflatedLog and the semantic convention that we defined there that we could add here?

silvadenisaraujo (commented):
Great to see this work materializing 🎉 added a few comments/questions.

src/service/logger/logger.ts
```diff
@@ -86,11 +87,34 @@ export class Logger {
...
+            this.logClient.error(logMessage, inflatedLog);
+            break;
+          default:
+            this.logClient.info(logMessage, inflatedLog);
+        }
+      } catch (e) {
+        console.error('Error using diagnostics client for logging:', e);
+      }
+    }
+    console.log(typeof inflatedLog === 'string' ? JSON.stringify(inflatedLog) : inflatedLog)
```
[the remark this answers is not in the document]
daniyelnnr: That's the current behavior, which logs data to stdout - just added some sanitization. I could've done this using diagnostics stdout exporter but, as @nicacioliveira said, that's not recommended since its use is for debugging purposes.

But the main question here is: do we still want to log data to stdout? Since our logs will be pushed to o11y collectors we could remove this feature as we may not want to support it.

What do you think @silvadenisaraujo?

src/service/telemetry/client.ts
daniyelnnr: As far as I checked, if APP.ID is unavailable we have some scenarios regarding an app using node-vtex-api:

- The app crashes because APP.ID is used in other parts of the code (e.g. current tracing initialization)
- Isolating the problem to telemetry client initialization only, we can observe:
  - Client creation succeeds due to the string `vtex-app`, used as fallback
  - The app throws an error requiring `serviceName` to proper init the client and it does not start.
 
So I don't see a major risk here, we have a fallback and we are throwing an error, I believe this is ensuring that we avoid starting the client with improper configurations.

src/service/logger/logger.ts
daniyelnnr: You're right we could've used it here and I didn't realise 😅
But taking a look at it, my first impression is that updating semantic convention as a whole for the project is a separate work from this. I thought about creating another PR to carry out this new work, do you think it's valid?

src/service/logger/client.ts
```diff
@@ -0,0 +1,54 @@
...
+    return initPromise;
+  }
+
+  isInitializing = true;
+  initPromise = initializeClient(account, workspace);
+
+  return initPromise;
+}
+
+async function initializeClient(account: string, workspace: string): Promise<LogClient> {
```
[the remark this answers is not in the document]
daniyelnnr: Commit f33bc589a57262e66c06001fb98393c139cb4ee7 added app name to log client initialization!

daniyelnnr:
> Would you mind adding a high-level overview of the changes? I had some trouble understanding the logging architecture here. Maybe I'll use copilot for this, but I think it would be beneficial if I could picture how this is organized

@arturpimentel I followed your idea and used copilot to get an overview of the logging architecture - if you think it makes sense, I can add a summary of all the changes as well.

## Original Logging Architecture

1. Proprietary logger with a specific VTEX format
2. Logs treated as separate entities from metrics and traces
3. Logs sent directly to stdout/stderr

```
[Application] → [VTEX Logger] → [stdout/stderr] → [Fluent bit] → [Log Collector]
```

## Architecture after the changes

1. Logs follow the OpenTelemetry standard format
2. Logs automatically correlated with traces and spans
3. Logs exported via the OTLP standard
4. Centralization of exporter configuration
5. Context information and metadata propagated automatically.
6. Logs automatically enriched with trace IDs and span IDs.

```
[Application] → [Diagnostics Logger] → [OpenTelemetry SDK] → [OTLP Exporter] → [Telemetry Backend/Collector]
```

Here is a more detailed overview:

```mermaid
flowchart LR
    A[Application Code] --[Logger Instance]--> B["Diagnostics Logger /src/service/logger/client.ts"]
    M[Trace/Span Context] -.-> B
    Z["Diagnostics Metrics [TBD]"] -.-> X
    Y["Diagnostics Traces [TBD]"] -.-> X
    
    B --> X
    X[Telemetry Client]--> C["OpenTelemetry Logs (@opentelemetry/sdk-logs)"]
    D["Resource & Attributes (Context enrichment)"] --> C
    
    
    E --> F["o11y Collector
    (io-telemetry-http.vtex.systems)"]
    
    subgraph Unified Telemetry
        C --> E["OTLP Exporter (gRPC/HTTP)"]
        C --> P["stdout Exporter (gRPC/HTTP)"]
        G["OpenTelemetry Metrics (@opentelemetry/sdk-metrics)"] --> E
        G["OpenTelemetry Metrics (@opentelemetry/sdk-metrics)"] --> P
        H["OpenTelemetry Traces (@opentelemetry/sdk-trace)"] --> E
        H["OpenTelemetry Traces (@opentelemetry/sdk-trace)"] --> P
    end
    
```

src/service/logger/client.ts
daniyelnnr: The CreateLogsExporterConfig function receives an object with some content on it to generate the exporter config, but all these fields are optional as we can see in the function signature - and internally it can handle the scenario where an endpoint or any other field is unavailable.

So if we have problems with the environment variable and/or don't proper inform the collector endpoint, the worst case scenario we'll get an functional exporter that sends logs to an invalid collector due to misconfiguration.

But yet, I'm not sure how I should proper inform the endpoint of the collector here.

src/caches/LRUDiskCache.ts
```diff
@@ -79,7 +79,7 @@ export class LRUDiskCache<V> implements CacheLayer<string, V>{
           resolve(fileData)
         } catch (e) {
           release()
-          resolve(undefined)
+          resolve(null as unknown as V)
```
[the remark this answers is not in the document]
daniyelnnr: This suggestion generates an error, I will keep it as it is

src/service/logger/logger.ts
silvadenisaraujo: I agree that's a valid approach :)

src/service/logger/client.ts
arturpimentel: I believe I wouldn't have a hard-coded default endpoint, since—as you showed—the `CreateLogsExporterConfig` can handle the `undefined` case. That exposes internal information in a public lib and also creates a config that is hard to modify: we need to redeploy the runtime if want to change it.

In summary, I would only get rid of the endpoint and let it be `process.env.OTEL_EXPORTER_OTPL_ENDPOINT`

src/service/logger/logger.ts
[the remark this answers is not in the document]
silvadenisaraujo: Discussion thread: https://vtex.slack.com/archives/C07V2SF5EDD/p1744132458802749

src/service/logger/client.ts
daniyelnnr: @arturpimentel I agree 100% about removing the hard-coded endpoint and will commit a change to fix this. However, I'm not sure where I could define the `process.env.OTEL_EXPORTER_OTLP_ENDPOINT`, as we don't want to set this in this lib. I imagine this info need to be defined somewhere else - is [service-runtime-node](https://github.com/vtex/service-runtime-node) the right place to do it?

src/service/logger/logger.ts
```diff
@@ -50,6 +32,25 @@ export class Logger {
         traceId: ctx.tracer.traceId,
       }
     }
+
+    this.initLogClient();
```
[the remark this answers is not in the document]
daniyelnnr: Ok, so we have the race condition problem while the logClient has not yet been initialized, which can lead to logs not being sent to the collector during this period. Based on the implementation of this PR and considering most of the time the initialization time should be short, we would have a small sample not being processed if logs are emitted right after the application starts.
I thought of two ways to approach this problem, doing (almost) nothing or implementing a log queue.

1. In a scenario where the logClient is not quickly initialized this can become a bigger problem since several logs can be lost - and an alternative mitigation would be to **apply a timeout to throw an error if it takes longer than a threshold to initialize**. The scenario where retrieving the logClient produces an error is already covered and I will not go into much detail about it. So for this approach we could solve it in a simplified way or even by doing nothing, assuming that the loss of initial logs is not a big problem.
2. If losing logs regardless of the quantity is a problem, we can follow a path of **queuing logs until the logClient is ready** - and when that happens they will be processed in the correct order. This approach is more reliable, but introduces some additional complexity.

WDYT @arturpimentel @silvadenisaraujo?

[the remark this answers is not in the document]
silvadenisaraujo: @daniyelnnr I completely agree with the option to **apply a timeout to throw an error if it takes longer than a threshold to initialize**, and I see that the best scenario would be to apply both (timeout + queue) but I wonder if it isn't an overkill we would need to consider a limit for this queue otherwise it would grow indefinitely, retries and so on. 

> So for this approach we could solve it in a simplified way or even by doing nothing, assuming that the loss of initial logs is not a big problem.

I'm assuming here that the logger will be a console.log() where the init is fast, and thinking the way that apps works as middlewares, it means that they don't have critical start up information, so I would be fine missing a few lines at start.

**The most important thing is that we need to ensure that the app is able to run even if the logger is not working properly.**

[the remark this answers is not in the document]
daniyelnnr: https://github.com/vtex/node-vtex-api/pull/577/commits/f47812d9c27ff3efa69c6d8a60680d64a15c68ba fixes this by using a timeout when log client startup exceeds a 5 seconds threshold. Tried to think a solution with queue/buffer, but the complexity increased due to the various edge cases to be considered and I preferred to go a safer path. I also ensured that the application was functional regardless of the log client initialization.

src/constants.ts
```diff
@@ -72,3 +72,5 @@ export const PRODUCTION = process.env.VTEX_PRODUCTION === 'true'
 export const INSPECT_DEBUGGER_PORT = 5858
 
 export const cancellableMethods = new Set(['GET', 'OPTIONS', 'HEAD'])
+
+export const LOG_CLIENT_INIT_TIMEOUT = 5000
```
silvadenisaraujo, line 76: Could you please add the unit to the constant, like SEC, MS.

src/service/logger/client.ts
```diff
@@ -0,0 +1,54 @@
...
+async function initializeClient(account: string, workspace: string, appName: string): Promise<LogClient> {
+  try {
+    const telemetryClient = await getTelemetryClient();
+
+    const logsConfig = Exporters.CreateLogsExporterConfig({
+      endpoint: process.env.OTEL_EXPORTER_OTLP_ENDPOINT,
+      path: process.env.OTEL_EXPORTER_OTLP_PATH || '/v1/logs',
+      protocol: 'http',
+      interval: 5,
+      timeoutSeconds: 5,
```
silvadenisaraujo, line 34: Should we use the constant here?

src/service/logger/logger.ts
```diff
@@ -50,6 +31,34 @@ export class Logger {
         traceId: ctx.tracer.traceId,
       }
     }
+
+    // this.initLogClient();
```
silvadenisaraujo, line 35: Should we enable it?

daniyelnnr: I disabled it so we keep on with sending logs through our fallback with console.log for now. Do you think we can handle this differently?

src/service/logger/client.ts
daniyelnnr: I thought about using it here so we can define a default fallback, but if it is a bad practice or not recommended we can remove this value, no problem.

silvadenisaraujo (approved):
LGTM

PR #586, opened 2025-06-13
Add a whitelist of headers to add in trace

silvadenisaraujo:
# Description

As of now, when adding traces in the VTEX IO node apps, we include all the headers from the request and responses, as we can see in the previous version of `cloneAndSanitizeHeaders`, this makes us to add too many information for our spans reaching limits of columns in our databases.

So this pull request adds a whitelist of headers that will be added to the span that actually brings meaningful o11y.

# Testing
- [x] Release a beta version
- [x] Upload to a testing cluster
- [x] Update image of the runtime with the beta version
- [x] Call apps forcing the trace and check if spans are created

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [6.49.5] - 2025-07-21
+### Changed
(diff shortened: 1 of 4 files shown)


arturpimentel (approved):
I believe this is pretty comprehensive for now

nicacioliveira (approved):
thank you 🙏🏼

PR #594, opened 2025-08-01
New auth scope

iagoaraujo:
Create the scope parameter in the Auth directive. In node-vtex-api, this parameter is optional, since the idea is that the same version of node-vtex-api will be used for GraphQL 1.x and 2.x. Builder-hub will validate the directive during the build when GraphQL is 2.x.

We're working on a vulnerability related to the lack of parameters indicating whether the query or mutation scope is public or private. I created the directive to accept both scopes, and there are two scenarios in which we need to run the authentication flow:
- when the scope is private
- when the scope is not defined, in which case it is private by default. In GraphQL 1.x, all those using the auth directive will be private without the "scope", so there is no compatibility break.

Link this branch into graphql apps and try to use it with graphql builders 1.x and 2.x (remember to link builder-hub with support to graphql 2.x).

#### Types of changes

### DIFF
--- a/src/service/worker/runtime/graphql/schema/schemaDirectives/Auth.ts
+++ b/src/service/worker/runtime/graphql/schema/schemaDirectives/Auth.ts
@@ -5,15 +5,10 @@ import { SchemaDirectiveVisitor } from 'graphql-tools'
 
 import { ServiceContext } from '../../../typings'
 
-enum Scope {
-  PUBLIC = 'PUBLIC',
-  PRIVATE = 'PRIVATE'
-}
-
 interface AuthDirectiveArgs {
   readonly productCode: string
   readonly resourceCode: string
-  readonly scope: Scope
+  readonly scope: 'PRIVATE' | 'PUBLIC'
 }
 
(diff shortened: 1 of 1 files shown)


daniyelnnr (approved):
LGTM!

src/service/worker/runtime/graphql/schema/schemaDirectives/Auth.ts
```diff
@@ -79,14 +84,23 @@ export class Auth extends SchemaDirectiveVisitor {
     const {resolve = defaultFieldResolver} = field
     field.resolve = async (root, args, ctx, info) => {
       const authArgs = parseArgs(this.args as AuthDirectiveArgs)
-      await auth(ctx, authArgs)
+      if (!authArgs.scope || authArgs.scope == 'PRIVATE') {
+        await auth(ctx, authArgs)
+      }
```
juliobguedes, line 89: Is this a breaking change? As far as I remember, we currently use public by default...

juliobguedes (approved):

PR #579, opened 2025-04-29
Add diagnostics-semconv to standardize some attributes and headers

daniyelnnr:
#### What is the purpose of this pull request?

This pull request introduces a major refactor to centralize HTTP header constants into a new `HeaderKeys` object within the `src/constants.ts` file. This change improves maintainability and reduces redundancy by replacing individual header constants with a single structured object. Additionally, it introduces a new `AttributeKeys` object for semantic attributes and updates the logger to use these attributes.

##### Dependency Additions
* Added a new dependency, `@vtex/diagnostics-semconv`, to support semantic conventions for diagnostics and tracing attributes.

##### Logger Enhancements
* Updated the `Logger` class to use the new `AttributeKeys` object for semantic attributes, such as `VTEX_OPERATION_ID` and `VTEX_ACCOUNT_NAME`, instead of plain object properties. This ensures better alignment with the diagnostics library.

##### Cleanup and Refactoring
* Removed unused imports and constants that were replaced by the centralized `HeaderKeys` object, further simplifying the codebase.

#### What problem is this solving?

Introduces `@vtex/diagnostics-semconv` library to support VTEX semantic conventions, in a reduced scope with respect to telemetry signals and emphasizing the context of logs. Future updates will include updates to the instrumentation using the diagnostics library, as well as the use of the semantic conventions of these signals.

#### How should this be manually tested?
The changes do not include new features or major behavior changes, so to test the changes simply build from this branch.

#### Types of changes

* [ ] Bug fix (a non-breaking change which fixes an issue)
* [x] New feature (a non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Requires change to documentation, which has been updated accordingly.

PR #611, opened 2025-09-26
fix: separate console and diagnostics client logging formats

daniyelnnr:

This PR fixes wrong mapping for some attributes, leading to fluent-bit being unable to create some attributes on OpenSearch, even though logs objects were being sent.

#579 introduced a small bug in log attribute mapping. Attributes like "account," "workspace," and "appId" were present in the object exported to stdout but weren't being mapped correctly in the OpenSearch attributes view. An [example](https://developer.logs.vtex.com/_dashboards/goto/ad14fec4769c3029155f3083fad2407d?security_tenant=global) is the image below, where we see a gap in the number of logs when filtering by one of the attributes with the affected mapping (you can also see that the issue was reversed after the rollback).

image

This issue was resolved by separating the object exported to stdout and logged to OpenSearch from the object logged by the diagnostics library client.

#### Types of changes

### DIFF
--- a/src/service/logger/logger.ts
+++ b/src/service/logger/logger.ts
@@ -79,17 +79,21 @@ export class Logger {
     cleanLog(data)
 
     /* tslint:disable:object-literal-sort-keys */
-    const inflatedLog = {
+    const commonLogFields = {
       __VTEX_IO_LOG: true,
       level,
+      operationId: this.operationId,
+      requestId: this.requestId,
+      ... (this.tracingState?.isTraceSampled ? { traceId: this.tracingState.traceId } : null),
+    }
+
+    const inflatedLog = {
       app,
       account: this.account,
       workspace: this.workspace,
       production: this.production,
-      operationId: this.operationId,
-      requestId: this.requestId,
       data,
-      ... (this.tracingState?.isTraceSampled ? { traceId: this.tracingState.traceId } : null),
+      ...commonLogFields,
     }
 
     // Mark third-party apps logs to send to skidder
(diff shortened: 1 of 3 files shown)


src/service/logger/logger.ts
```diff
@@ -101,30 +100,45 @@ export class Logger {
...
+      __VTEX_IO_LOG: true,
+      level,
+      [AttributeKeys.VTEX_IO_APP_ID]: app,
+      [AttributeKeys.VTEX_ACCOUNT_NAME]: this.account,
+      [AttributeKeys.VTEX_IO_WORKSPACE_NAME]: this.workspace,
+      [AttributeKeys.VTEX_IO_WORKSPACE_TYPE]: this.production ? 'production' : 'development',
+      [AttributeKeys.VTEX_IO_APP_AUTHOR_TYPE]: APP.IS_THIRD_PARTY() ? '3p' : '1p',
+      operationId: this.operationId,
+      requestId: this.requestId,
+      data,
```
juliobguedes, line 115: Remember to remove `data` from this log to avoid duplications, as we talked yesterday.

src/service/logger/logger.ts
```diff
@@ -82,11 +82,10 @@ export class Logger {
     const inflatedLog = {
```
juliobguedes, line 82: Suggestion: have an object called `commonLogFields` which sets `__VTEX_IO_LOG`, `level`, `operationId`, `requestId`, `data`, and `traceId`. Having this object, define inflatedLogs and diagnosticsLogs as

```js
const inflatedLogs = { ...commonLogFields, app, account: '...' }; // ... other fields
const diagnosticsLogs = { ...commonLogFields, [AttributeKeys.VTEX_IO_APP_ID]: app }; // ... other fields
```

WDYT?

silvadenisaraujo (approved):
Agree with comments from @juliobguedes , besides LGTM!

PR #667, opened 2026-07-16
Aggregate prom-client metrics across cluster workers for /metrics

silvadenisaraujo:
## Summary

VTEX IO runtimes run as a Node.js cluster of `min(cpus, 4)` workers, each keeping its own prom-client registry, and `/metrics` is answered by whichever worker the shared-port round-robin hands the scrape to. Because keep-alive (5s) is shorter than the scrape interval, nearly every scrape hits a different worker, so Prometheus stores an interleaved braid of 4 independent counters under one instance label — treating every downward sample as a counter reset and inflating `increase()`/`rate()` on `runtime_*` counters by orders of magnitude (~×40 measured live). This PR fixes that by aggregating across workers with prom-client's `AggregatorRegistry` (pinned `^14.2.0`): the master builds a single merged, monotonic view over the existing cluster IPC and the worker answering the scrape serves that aggregate. Single-worker / LINKED mode stays unchanged.

## Spec

Accepted. See [`specs/aggregate-prom-client-metrics-across-cluster-wor.md`](specs/aggregate-prom-client-metrics-across-cluster-wor.md).

## Implementation

- `src/service/metrics/clusterMetricsAggregator.ts` (new): the master-side `AggregatorRegistry` plus a small tagged-IPC protocol. A worker answering a `/metrics` scrape asks the master over the existing cluster IPC (`AGG_METRICS_REQ`), the master builds the merged, monotonic cluster view via `clusterMetrics()` and replies (`AGG_METRICS_RES`). prom-client's own `prom-client:*` cluster messages are recognised and ignored by our handlers. On timeout/error the worker falls back to its local `register.metrics()` so `/metrics` never hangs.
- `src/service/master.ts`: initialises the master aggregator when `workers > 1` and routes worker aggregate requests / prom-client messages in `onMessage`.
- `src/service/worker/index.ts`: installs the worker-side cluster responder when `workers > 1` and routes the master's aggregate response in `onMessage`.
- `src/service/worker/runtime/builtIn/middlewares.ts`: `prometheusLoggerMiddleware(workers)` serves the cluster aggregate in multi-worker mode and the local default registry in single-worker mode. Scrapes are still excluded from request counters (early return preserved) and metric names/labels are unchanged.

Covered by tests proving monotonicity across whichever worker answers, sum-across-N-workers, single-worker default-registry behaviour, and presence of default process metrics.

## Testing

- Beta tag: v7.4.1-beta.X
                                                                         
## Validation

[...]

PR #673, opened 2026-08-03
fix(metrics): never emit runtime_http_* samples without a handler label

silvadenisaraujo:
## Problem

After `/metrics` started serving the cluster-wide aggregate (#667, `@vtex/api@7.4.1`, `service-node:7.7.14`), dashboards grew an extra, nameless line:

```
runtime_http_requests_total{handler="builtin:healthcheck",status_code="200"} 471
runtime_http_requests_total{handler="private-handler:ssr",status_code="200"} 421
runtime_http_requests_total{status_code="200"} 21   ← no handler label at all
runtime_http_requests_total{status_code="404"} 2    ← no handler label at all
```

Observed live on `vtex-render-ssr` in `prod-dj-ioadmin-eks-use1a-t1d` / `vendor-vtex`.

image

### Root cause

`ctx.requestHandlerName` is only assigned inside a route pipeline (`nameSpanOperationMiddleware`) or by a builtin handler, but `addRequestMetricsMiddleware` is mounted at the top of the chain (`worker/index.ts:245`) and counts **every** request in a `finally` block. So requests that never reach a named handler are counted with `handler: undefined`.

prom-client keeps the key in memory, so the local exposition rendered it as `handler="undefined"`. **Node's cluster IPC serializes messages as JSON, and `JSON.stringify` drops `undefined` values**, so the sample reaches the master with the `handler` key gone. Verified against the pinned `prom-client@14.2.0`:

```
stored labels object keys: [ 'handler', 'status_code' ]  has handler prop: true  value: undefined

local registry (pre-aggregation / workers===1):  m_total{handler="undefined",status_code="200"} 2
aggregate WITHOUT ipc serialization:             m_total{handler="undefined",status_code="200"} 2
aggregate AFTER  ipc serialization (real path):  m_total{status_code="200"} 2      ← label stripped
```

Prometheus reads an absent label as `handler=""`, so:

- it is a **different series** from the historical `handler="undefined"` → panels split at the rollout boundary;
- Grafana has no value to interpolate into `{{handler}}` and falls back to the default field name `Value`;
- exclusion filters stop working — `handler!~"builtin:.*|undefined"` does not match `""`, so the bucket that used to be filtered out is now included.

### Which requests are affected

| source | status | note |
| --- | --- | --- |
| `GET /_status` (platform status poller) | 200 | reaches `statusTrackHandler`, which sets only the **span** name, never `ctx.requestHandlerName` |
| unmatched paths | 404 | no pvt route matches and no `x-colossus-route-id` → chain ends, Koa answers its default 404 |
| replica-level rate limit | 429 | `concurrentRateLimiter` is mounted before the routers and throws |
| unimplemented / unknown route id | 501 / 404 / 400 | `routerFromPublicHttpHandlers`, `routerFromEventHandlers` return before naming |
| aborted requests | — | `abortedRequests.inc` uses the same undefined name |

Reproduced live: 5× `GET /_status` moved `{status_code="200"}` by exactly +5 (+1 background poll); two requests to unmatched paths created `{status_code="404"} 2`; `HEAD /healthcheck` and `GET /_metrics` stayed correctly labelled as `builtin:healthcheck` / `builtin:metrics-logger`.

## Proposal

1. **`src/service/metrics/requestHandlerLabel.ts` (new)** — one place that resolves the label, falling back to `'undefined'`, with the reasoning documented next to the constant.

   `'undefined'` rather than a nicer word like `'unnamed'` is deliberate: it is exactly what prom-client rendered locally before cluster aggregation existed, so the aggregated output keeps the historical series identity and dashboards/alerts already filtering on `handler="undefined"` (e.g. `handler!~"builtin:.*|undefined"`) keep working with no query changes. Empty strings fall back too, so the label is never emitted empty.

2. **`requestMetricsMiddleware.ts` / `otelRequestMetricsMiddleware.ts`** — use the helper at all four call sites each (total, aborted, response sizes, timings). Evaluation stays inside the callbacks/`finally`, so the handler name is still read *after* the pipeline ran.

3. **`statusTrack.ts`** — set `ctx.requestHandlerName = 'builtin:status-track'`, parity with the three sibling builtins. `/_status` traffic gets its own series instead of polluting the catch-all bucket. This commit is separable if reviewers prefer to ship only (1)+(2) — note `/_status` is in `PATHS_BLACKLISTED_FOR_TRACING`, so the pre-existing `setOperationName` call is usually a no-op, which is likely why the missing assignment went unnoticed.

No metric names, help text, buckets or label *names* change.

## Tests

- `src/service/metrics/__tests__/requestHandlerLabel.test.ts` (new) — drives the real middleware and asserts the label survives a **cluster IPC JSON round-trip**, that no sample is emitted with a missing/empty `handler`, that named and unnamed handlers stay separate series, and that aborted requests are labelled. Reverting the fallback makes 5 of these 7 cases fail.
- `src/service/metrics/__tests__/clusterMetricsAggregator.test.ts` — the aggregation helper now round-trips worker registries through JSON, so these tests exercise what the master actually receives. The absence of that round-trip is why #667 didn't catch this.
- `src/service/worker/runtime/__tests__/statusTrack.test.ts` (new) — asserts `/_status` names itself, with and without tracing.

`jest`: 17 suites, 239 passed (24 pre-existing skips). `tsc --noEmit` clean. `tslint` reports no new findings.

## Rollout note

The unnamed series (`handler=""`) exists only on runtimes carrying #667 without this fix, i.e. `service-node:7.7.14` up to the release that includes this PR. Dashboards looking back across that window can stitch the two shapes with:

```promql
sum by (handler) (
  label_replace(
    rate(runtime_http_requests_total{cluster=~"$cluster", app="$app_name"}[$__rate_interval]),
    "handler", "undefined", "handler", "^$"
  )
)
```

Do the `label_replace` **inside** the aggregation, otherwise the relabelled series can collide with a real `handler="undefined"` series during the rollout and Prometheus errors with `vector cannot contain metrics with the same labelset`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [7.5.0]
+
+### Added
+
+- Kubernetes cluster identity and role (`VTEX_CLUSTER_ID` → `cluster_id`,
+  `VTEX_CLUSTER_ROLE` → `cluster_role`) as shared diagnostics resource
+  attributes, applied centrally to diagnostics metrics and logs. Values are
+  trimmed and omitted independently when empty, enabling cross-cluster
+  dashboards and incident isolation without consuming per-call custom
+  attribute limits.
+
 ### Fixed
+
 - `runtime_http_*` metrics no longer emit samples without a `handler` label. Requests
   that never reach a named handler (unmatched paths, replica-level rate limit
   rejections, errors before the route pipeline) were counted with
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- `runtime_http_*` metrics no longer emit samples without a `handler` label. Requests
+  that never reach a named handler (unmatched paths, replica-level rate limit
+  rejections, errors before the route pipeline) were counted with
+  `handler: undefined`; Node's cluster IPC serializes worker registries as JSON, which
+  drops `undefined` values, so the aggregated `/metrics` exposed a second, unnamed
+  series that Prometheus reads as `handler=""`. Those requests are now labelled
+  `handler="undefined"` — the same value prom-client rendered locally before cluster
+  aggregation — keeping dashboards and alerts that filter on it working.
+- `/_status` requests are now reported as `handler="builtin:status-track"`, matching
+  the other builtin handlers, instead of falling into the unnamed bucket.
 
 ## [7.4.0] - 2026-06-22
 ### Changed
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@vtex/api",
-  "version": "7.4.2",
+  "version": "7.5.0",
   "description": "VTEX I/O API client",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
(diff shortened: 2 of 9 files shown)


caroolcanelas2 (approved):

juliobguedes (approved):

juliobguedes:
Ignoring SonarQube as its comments are not related to this change, but part of our technical debts in the backlog.

PR #676, opened 2026-08-03
refactor(metrics): drop metric/status log export

iagoaraujo:
## Summary

- Removes the `metric/status` log line (`logStatus`) that `trackStatus()` wrote to stdout, along with the `ACCOUNT`, `APP`, `PRODUCTION` and `WORKSPACE` imports that became unused.
- Keeps the `global.metrics.statusTrack()` call even though its return value is now discarded. It delegates to `flushMetrics()`, which resets the metric accumulators, updates the CPU usage baseline and clears the incoming request stats — dropping the call would let the accumulators grow unbounded and turn the CPU/request metrics into cumulative values instead of deltas.
- The status-track flow itself is untouched: `HttpAgentSingleton.updateHttpAgentMetrics()`, the `/_status` handler and the master/worker broadcast all behave as before.

Companion change in `service-runtime-base` (branch `chore/remove-metric-status-log`) removes the equivalent log emitted by the Go runtime.

## Test plan

- [x] `yarn lint` — clean (only pre-existing warnings)
- [x] `yarn build`
- [x] `yarn test` — 11 suites, 206 tests passing

Made with [Cursor](https://cursor.com)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@vtex/api",
-  "version": "7.4.0",
+  "version": "7.4.2-beta.1",
   "description": "VTEX I/O API client",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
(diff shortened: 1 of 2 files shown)


silvadenisaraujo (approved):
nice :)

daniyelnnr (approved):

PR #713, opened 2026-09-21
fix(HttpClient): hash Buffer bodies directly in getWithBody instead of via JSON.stringify

mendescamara:
## Problem

`getWithBody` always computes its cache-key `bodyHash` with `createHash('md5').update(JSON.stringify(data, deterministicReplacer))`, even when `data` is a `Buffer`. Consumers that compress the body (e.g. gzip) before calling `getWithBody` pass the resulting `Buffer` as `data` — this is exactly what render-ssr does for bodies over 1KB (`node/clients/renderSSR.ts:19-27`), which is what surfaced this while investigating vtex/render-server#859. This isn't specific to render-server: any consumer that passes a compressed `Buffer` to `getWithBody` hits the same issue.

When `data` is a `Buffer`:

1. **Correctness/cache**: `JSON.stringify` invokes `Buffer.prototype.toJSON()`, which serializes it as `{"type":"Buffer","data":[...]}` — the byte array written as decimal text. The resulting hash reflects that transport representation, not the actual bytes/content, so the cache-key query param used by KubeRouter is derived from something other than the real payload.
2. **Performance**: serializing a large buffer's JSON representation is far more expensive than hashing its bytes directly.

## Fix

Extracted the hash computation to `src/utils/bodyHash.ts` (`computeBodyHash`). When `Buffer.isBuffer(data)`, hash the buffer directly: `createHash('md5').update(data).digest('hex')`. For any other `data` type, behavior is unchanged (`JSON.stringify` + the existing deterministic key-sorting replacer).

## Compatibility

This changes the `bodyHash` value (and therefore the cache-key query param) for any existing consumer that passes a `Buffer` to `getWithBody`. Effect is safe: worst case is a one-time cache miss on stale KubeRouter entries, not incorrect content being served across different payloads. Documented in `CHANGELOG.md` under `Unreleased`.

## Testing

`src/utils/bodyHash.test.ts` (TDD — written and run RED before implementing):

- two buffers with identical bytes → same hash
- buffers with different bytes → different hashes
- buffer hash differs from what the old `JSON.stringify` approach would produce
- non-`Buffer` (plain object) hash is unchanged from current behavior, regardless of key order
- the same JSON payload, gzip-compressed independently twice (simulating two separate requests, not the same `Buffer` instance) produces byte-identical buffers and identical hashes — validates that caching still works across requests for the render-ssr scenario
- gzip buffers from different underlying JSON payloads produce different hashes

Full suite: 20/20 suites, 253/253 tests passing. `tsc --noEmit` clean; `tslint` on touched files shows no new warnings.

## Benchmark (informal, local, using the real `computeBodyHash`)

| Buffer size | Before (`JSON.stringify`) | After (direct hash) |
|---|---|---|
| 50 KB | ~4.1 ms | ~0.06 ms |
| 200 KB | ~18.0 ms | ~0.21 ms |
| 1 MB | ~88.9 ms | ~1.01 ms |

Out of scope: render-server/render-ssr changes — this fix is isolated to `@vtex/api`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/utils/bodyHash.ts
```diff
@@ -0,0 +1,33 @@
...
+      // I don't believe this will ever happen, but just in case
+      // Also, I didn't include error as I am unsure if it would have sensitive information
+      onSerializeError?.()
+      return value
+    }
+  }
+
+  // MD5 here only derives a cache-key digest for the request body, not a security-sensitive
+  // value - no secret protection or tamper-integrity guarantee is being made.
+  return createHash('md5').update(JSON.stringify(data, replacer)).digest('hex') // NOSONAR
```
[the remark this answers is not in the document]
mendescamara: Fixed in f6277e23 (`fix(bodyHash): never throw from computeBodyHash, even on serialization failure`). Wrapped the whole `JSON.stringify(data, replacer)` + digest step in its own try/catch, so the second `Get` that `JSON.stringify` performs outside the replacer is now covered too. Added a test with a getter that throws on every access (not just the first) to lock this in.

src/utils/bodyHash.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  const replacer = (key: string, value: any) => {
+    try {
+      return deterministicReplacer(key, value)
+    }
+    catch {
+      // I don't believe this will ever happen, but just in case
+      // Also, I didn't include error as I am unsure if it would have sensitive information
+      onSerializeError?.()
```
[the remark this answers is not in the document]
mendescamara: Fixed in f6277e23 (`fix(bodyHash): never throw from computeBodyHash, even on serialization failure`). The outer try/catch around `JSON.stringify` + digest now covers this too: `JSON.stringify(undefined, replacer)` returns `undefined`, `Hash.update(undefined)` throws, and that's now caught by the same guard. Added a unit test for `computeBodyHash(undefined)` (with and without an `onSerializeError` callback).

src/utils/bodyHash.ts
```diff
@@ -0,0 +1,43 @@
...
+    return createHash('md5').update(JSON.stringify(data, replacer)).digest('hex') // NOSONAR
+  }
+  catch {
+    // JSON.stringify can still fail (or return undefined) even after the replacer recovers -
+    // e.g. a property whose getter fails on every access (not just the one the replacer
+    // already caught), or `data` itself serializing to `undefined` (e.g. data === undefined).
+    // Fall back to a representation that never throws.
+    onSerializeError?.()
+    return createHash('md5').update(String(data)).digest('hex') // NOSONAR
+  }
```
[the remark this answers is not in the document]
mendescamara: Fixed in 53e8508d (`fix(bodyHash): avoid cache-key collisions on the serialization-failure fallback`). Good catch - replaced the `String(data)` fallback with `randomBytes(16).toString('hex')`, so every call that hits this path gets a unique key instead of a constant one. This guarantees the fallback can only ever cause a cache miss, never a collision between different bodies, matching the CHANGELOG's stated worst case. Added a test with two structurally different circular-reference bodies asserting they no longer produce the same hash (it failed with the old `String(data)` fallback before this fix, confirming the collision).

src/utils/bodyHash.ts
```diff
@@ -0,0 +1,46 @@
...
+// Object keys from `Object.entries` are always distinct, so `ka` and `kb` are never equal here.
+const compareKeys = ([ka]: [string, unknown], [kb]: [string, unknown]) => (ka < kb ? -1 : 1)
+
+const deterministicReplacer = (_: any, v: any) => {
+  return typeof v !== 'object' || v === null || Array.isArray(v) ? v :
+    Object.fromEntries(Object.entries(v).sort(compareKeys))
+}
+
+export function computeBodyHash(data: any, onSerializeError?: () => void): string {
+  if (Buffer.isBuffer(data)) {
```
Sophmrs: nit: use `ArrayBuffer.isView(data)` instead as it still works for `Buffer` but also enables other types like `Uint8Array` to use this optimization, and hash.update() already hashes just the view's byte window correctly.
```suggestion
  if (ArrayBuffer.isView(data)) {
```

src/utils/bodyHash.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  const replacer = (key: string, value: any) => {
+    try {
+      return deterministicReplacer(key, value)
+    }
+    catch {
+      // I don't believe this will ever happen, but just in case
+      // Also, I didn't include error as I am unsure if it would have sensitive information
+      onSerializeError?.()
```
Sophmrs: Do we care that this could log duplicate errors for the same request if the serialization fails?

src/HttpClient/HttpClient.getWithBody.test.ts
Sophmrs: Might be worth adding a small test to assert that a caller-supplied `config.params[BODY_HASH]` always gets overridden by the computed hash

Sophmrs (approved):
Nice! Just left a couple of suggestions.

src/utils/bodyHash.ts
mendescamara: Fixed in 6b87f63 - switched to `ArrayBuffer.isView(data)`. This widens the fast-path to any TypedArray/DataView (not just `Buffer`), and `hash.update()` correctly hashes just the view's byte window as you noted. Added tests hashing a `Uint8Array` and confirming it matches the equivalent `Buffer`.

src/utils/bodyHash.ts
mendescamara: Good catch - yes, this could double-log for the same request (once from the replacer's catch, once from the outer JSON.stringify catch for the same underlying failure). Fixed in 6b87f63: `onSerializeError` is now deduplicated per `computeBodyHash` call via a local flag, so `logger.warn` fires at most once per request regardless of how many serialization steps observe the failure. Updated the `alwaysFlaky` test to assert `toHaveBeenCalledTimes(1)`.

src/HttpClient/HttpClient.getWithBody.test.ts
mendescamara: Added in 6b87f63: a new test asserting that a caller-supplied `config.params[BODY_HASH]` is always overridden by the computed hash, so this stays locked in against future refactors.

src/utils/bodyHash.ts
```diff
@@ -0,0 +1,59 @@
...
+    // JSON.stringify can still fail (or return undefined) even after the replacer recovers -
+    // e.g. a property whose getter fails on every access (not just the one the replacer
+    // already caught), a circular reference, or `data` itself serializing to `undefined`
+    // (e.g. data === undefined). A constant fallback (e.g. String(data)) would collapse
+    // any two different bodies that hit this path into the same bodyHash - a cache-key
+    // collision, not just a miss. Use random bytes instead: every call gets a unique key,
+    // so this path can only ever cause a cache miss, never serve the wrong content.
+    reportSerializeError()
+    return randomBytes(16).toString('hex')
+  }
```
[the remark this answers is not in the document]
mendescamara: Fixed in 49efb22 - added an explicit `data === undefined` check before the JSON.stringify try/catch, hashing a fixed sentinel (`'undefined'`) instead of falling through to the randomBytes fallback. This keeps the bodyHash/cache-key stable for callers that omit the body, while the randomBytes fallback stays reserved for genuine serialization failures (circular refs, always-throwing getters). Added a test asserting computeBodyHash(undefined) is stable across calls, and updated the existing undefined test to assert onSerializeError is no longer called for this case (it's not a failure anymore).

src/utils/bodyHash.ts
```diff
@@ -0,0 +1,67 @@
...
+      return value
+    }
+  }
+
+  try {
+    // MD5 here only derives a cache-key digest for the request body, not a security-sensitive
+    // value - no secret protection or tamper-integrity guarantee is being made.
+    return createHash('md5').update(JSON.stringify(data, replacer)).digest('hex') // NOSONAR
+  }
+  catch {
```
iago1501: Non-blocking: worth calling out explicitly (here or in the changelog) that this changes an error-handling contract, not just a performance/cache-key detail.

Previously, `getWithBody` had no outer try/catch around `JSON.stringify` — if `data` had a circular reference (or any other unserializable shape), `JSON.stringify` threw and the failure propagated to the caller. Now that failure path is caught here and silently downgraded to a `randomBytes` fallback (logged via `onSerializeError`, but not surfaced to the caller as an error).

That's a legitimate fail-fast → fail-open tradeoff (worst case is a cache-miss, never wrong content served, and it's well-reasoned in the comments above), but it's a behavior change independent of the Buffer-hashing fix this PR is about: a caller that was previously relying on the throw to catch a bug in the data it was passing (e.g. accidentally building a circular structure) will now silently get a working-but-uncached request instead of an exception.

Worth confirming this is intentional and maybe adding a one-line note about it alongside the existing Buffer-hash compatibility note in the changelog, so it's not just implicit in the code comments.

iago1501 (commented):
Nenhum achado bloqueante, daí já deixei o approve. Um comentário inline não-bloqueante abaixo sobre uma mudança de contrato de erro (fail-fast → fail-open) que vale confirmar como intencional.

iago1501 (approved):

src/utils/bodyHash.ts
mendescamara: Fair point, and confirmed intentional - added in d02c748: a new CHANGELOG entry explicitly calling out this fail-fast → fail-open change (previously an unhandled JSON.stringify throw, e.g. from a circular reference, would propagate out of getWithBody to the caller; now it's caught and downgraded to a one-off random bodyHash instead). Thanks for flagging it - it was only implicit in the code comments before.

Sophmrs: Cases that would fail json.stringify will end up failing afterwards in axios anyway, won't they?

mendescamara: Yes, because the payload is invalid!

Sophmrs: So the changelog is incorrect, right?

